PR #3, opened 2019-08-06
Improved the appearance of the graph and fixed code style.

Honny1:
Graph has now better position and edges has color by node form lower row.


dahaic:
Great job with the pep8 style changes :) In general  - don't be afraid to do more smaller pull requests. If you feel it's too much overhead, try to automate it a bit.

Right now, you are merging code yourself, but as soon as you will require code review from somebody, you want them to have really easy job. Otherwise you'll wait longer and whole process might get tedious. So it's better to get the habits right from the beginning.

jan-cerny:
@Honny1 Let's try it the next time! When you will have a new pull request you can request a review. I will be happy to give you a review :smile:

Honny1:
@dahaic  Tahanks, I will start make pull request with smaller count of commits.
@jan-cerny  As you wish. smile

PR #4, opened 2019-08-07
 Change the top node label to name of rule.

Honny1:


Honny1:
@jan-cerny If you have time. You can make review. :smile:

graph/oval_graph.py
```diff
@@ -161,75 +161,52 @@ def change_tree_value(self, node_id, value):
...
-                    "x": x,
-                    "y": y,
-                    "size": 3,
-                    "color": '#000000'
-                }
+                return '#000000'
+
+    def _get_node_title(self):
+        if self._get_color_node() == '#000000':
+            return str(self.node_id) + ' ' + self.value
```
jan-cerny: I don't like this condition. The condition means: If the node is black then add its value. However, the black color is only a consequence of a different property.

Even more, it is hard to know what is the meaning of the black color. I can't figure out when this condition is true without running the code and seeing the output. Is it the self.value is non-empty? Or when does it happen?

What if in future you decide to use a different color? Then you would have to remember that you have to change this code as well. I'm sure you will forget about it and then you will wonder why it doesn't work and you might spend a lot of time investigating it.

I believe you should make the behavior obvious for anyone who reads the code. Please think how to rewrite the condition so that it doesn't depend on a color but it directly checks some property of the class.

graph/oval_graph.py
```diff
@@ -161,75 +161,52 @@ def change_tree_value(self, node_id, value):
...
-                'id': self.node_id,
-                'label': self._get_label(),
-                'url': 'null',
-                'text': 'null',
-                'title': self.node_id,
-                "x": x,
-                "y": y,
-                "size": 3,
-                "color": '#00ff00'}
+            return '#00ff00'
```
jan-cerny: It would be better to define a dictionary that maps the value to a color.

For example:
```
VALUE_TO_COLOR = {
   "true" : "#00ff00",
   "false" : "#ff0000",
   "": "#000000"
}
```
And then use the dictionary in a function.


If you use the dictionary approach then all the colors would be defined at a single place, which is always considered better than having them spread across the code.

graph/oval_graph.py
```diff
@@ -161,75 +161,52 @@ def change_tree_value(self, node_id, value):
...
-                    'label': self.value,
-                    'url': 'null',
-                    'text': 'null',
-                    'title': self.node_id,
-                    "x": x,
-                    "y": y,
-                    "size": 3,
-                    "color": '#00ff00'
-                }
+                return '#00ff00'
```
jan-cerny: `self.evaluate_tree()` is called twice, which doesn't seem to be necessary, you can call it only once and save the returned value to a variable, and then use that variable in your conditions.

graph/oval_graph.py
```diff
@@ -378,9 +355,9 @@ def change_position(self, positions, nodes_in_rows):
 
         for row in nodes_in_rows:
             for node in nodes_in_rows[row]:
-                if len(
-                        node['label']) > 6 and len(
-                        node['label']) < 40 or continue_move:
+                if len(node['label']) > 6\
+                        and len(node['label']) < 40\
```
jan-cerny: Using the backslashes is ugly and I think that they're not necessary.
You can silence the PEP8 warning by enclosing the whole condition inside parentheses.

Something like this could work:
```
if (len(node['label']) > 6                                                     
        and len(node['label']) < 40                                            
        or continue_move):                                                     
    print("hi")  
```

Honny1:
@jan-cerny 
I fixed all code mistakes, but I hat problem with rebase. When I use `git rebase -i upstream / master`. Then, a vim window will pop up, show only two commit. I can't see there some merge commit". 

Example:
pick e4d2623 Change the top node label to name of rule.
pick 27b79b7 Change get_node_color() and get_node_title() and some PEP8 stuff

jan-cerny:
It's green :tada: :1st_place_medal:

PR #12, opened 2019-08-12
Negation in ovalNode

Honny1:
Now ovalNode can work with negation.

### DIFF
--- a/playground.py
+++ b/playground.py
@@ -1,9 +1,26 @@
 import graph.oval_graph
+import graph.xml_parser
+import json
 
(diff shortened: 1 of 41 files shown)


playground.py
```diff
@@ -0,0 +1,26 @@
+import graph.oval_graph
```
matejak: Not every file has to be part of the repository. But if it is, it should have some description (e.g. as comments), and if it is a script, a proper command-line interface would be great too.

matejak:
Thanks for the update! There are some dark corners in the code, but they can be addressed in another PRs.

PR #16, opened 2019-08-15
Fix loading negation from xml

Honny1:


matejak:
When I try to run the code, I get an error 

```
Traceback (most recent call last):
  File "run.py", line 35, in <module>
    arg.func(arg)
  File "run.py", line 23, in run
    with open('html_interpreter/data/data.js', "w+") as file:
IOError: [Errno 2] No such file or directory: 'html_interpreter/data/data.js'
```

The subdirectory `data` is really not there.

Issue #17, opened 2019-08-15
Notselected rules reported as not found.


matejak:
When there is an ARF that contains a definition of a rule, but the rule itself was not tested, the `run.py` will produce the `err- 404 rule not found!` error.

This leaves the user wondering whether they got the rule ID right, which is not the problem. A better message would be `no results found for rule ...`, but even better would be `rule ... was not selected, so there are no results`.

matejak:
I confirm that the issue is fixed, thanks!

PR #19, opened 2019-08-16
Recognition of notselected rules

Honny1:
Fix issue #17

### DIFF
--- a/graph/xml_parser.py
+++ b/graph/xml_parser.py
@@ -27,32 +27,33 @@ class xml_parser():
 
     def get_data(self, href):
         ns = {
-            'ns0': 'http://oval.mitre.org/XMLSchema/oval-results-5',
-            'ns1': 'http://scap.nist.gov/schema/asset-reporting-format/1.1'
+            'XMLSchema': 'http://oval.mitre.org/XMLSchema/oval-results-5',
+            'arf': 'http://scap.nist.gov/schema/asset-reporting-format/1.1'
         }
 
         report_data = None
-        reports = self.root.find('.//ns1:reports', ns)
+        reports = self.root.find('.//arf:reports', ns)
         for report in reports:
             if "#" + str(report.get("id")) == href:
                 report_data = report
 
         trees_data = report_data.find(
-            './/ns0:oval_results/ns0:results/ns0:system/ns0:definitions', ns)
+            ('.//XMLSchema:oval_results/XMLSchema:results/'
+             'XMLSchema:system/XMLSchema:definitions'), ns)
         return trees_data
 
     def get_used_rules(self):
         ns = {
-            'ns0': 'http://checklists.nist.gov/xccdf/1.2',
+            'xccdf': 'http://checklists.nist.gov/xccdf/1.2',
         }
         rulesResults = self.root.findall(
-            './/ns0:TestResult/ns0:rule-result', ns)
+            './/xccdf:TestResult/xccdf:rule-result', ns)
         rules = []
         for ruleResult in rulesResults:
-            result = ruleResult.find('.//ns0:result', ns)
+            result = ruleResult.find('.//xccdf:result', ns)
             if result.text != "notselected":
                 check_content_ref = ruleResult.find(
-                    './/ns0:check/ns0:check-content-ref', ns)
+                    './/xccdf:check/xccdf:check-content-ref', ns)
                 if check_content_ref is not None:
                     rules.append(dict(
                         id_rule=ruleResult.get('idref'),
@@ -63,13 +64,13 @@ class xml_parser():
 
     def get_notselected_rules(self):
         ns = {
-            'ns0': 'http://checklists.nist.gov/xccdf/1.2',
+            'xccdf': 'http://checklists.nist.gov/xccdf/1.2',
         }
         rulesResults = self.root.findall(
-            './/ns0:TestResult/ns0:rule-result', ns)
+            './/xccdf:TestResult/xccdf:rule-result', ns)
         rules = []
         for ruleResult in rulesResults:
-            result = ruleResult.find('.//ns0:result', ns)
+            result = ruleResult.find('.//xccdf:result', ns)
             if result.text == "notselected":
                 rules.append(dict(id_rule=ruleResult.get('idref')))
         return rules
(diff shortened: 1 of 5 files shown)


tests/test_sigmaJs.py
```diff
@@ -361,17 +362,40 @@ def test_get_def_id_by_rule_id():
...
 
 
+def test_get_def_id_by_notselected_rule_id():
+    src = 'test_data/ssg-fedora-ds-arf.xml'
+
+    parser = tests.any_test_help.get_parser(src)
+    rule_id = 'xccdf_org.ssgproject.content_rule_ntpd_specify_remote_server'
+    with pytest.raises(ValueError) as e:
+        parser.get_def_id_by_rule_id(rule_id)
+    assert str(
```
matejak: `str(e)` is better than `e.value`, as usage of the `value` attribute is not [documented](https://docs.python.org/3.7/library/exceptions.html#BaseException), while `str()` is.

tests/test_sigmaJs.py
```diff
@@ -361,17 +362,40 @@ def test_get_def_id_by_rule_id():
...
+        e.value) == ('err- rule "' +
```
matejak: You can get the string on one line using the formatted string literals.

tests/test_sigmaJs.py
```diff
@@ -356,23 +356,18 @@ def test_get_def_id_by_rule_id():
...
     parser = tests.any_test_help.get_parser(src)
     rule_id = 'xccdf_org.ssgproject.content_rule_ntpd_specify_remote_server'
-    with pytest.raises(ValueError) as e:
-        parser.get_def_id_by_rule_id(rule_id)
-    assert str(
-        e.value) == ('err- rule "' +
-                     rule_id +
-                     '" was not selected, so there are no results.')
+
+    with pytest.raises(Exception, match="err- rule \"{}\" was not selected, so there are no results.".format(rule_id)):
```
matejak: Very clever use of the `match` argument!
I have those small suggestions:


- Use single quote for quoting values, it is the common pattern. And try to avoid escaping quotes - you don't have to do it in Python (you can use `"`, `'`, or even `"""`, `'''` for the outer string to avoid quote escaping). This is probably an issue in the source code, not in the test.
- Make the matches more relaxed, so the tests won't break when you decide to make minor modifications to the error message. I would just check for the presence of the rule ID and the `not selected` phrase.

tests/test_sigmaJs.py
```diff
@@ -381,21 +376,17 @@ def test_str_to_bool():
...
 def test_use_bat_report_file():
     src = ('test_data/xccdf_org.ssgproject.content_rule_sssd_' +
            'ssh_known_hosts_timeout-comment.fail.sh-xccdf_or' +
            'g.ssgproject.content_profile_ospp-results-initial.xml')
 
-    with pytest.raises(ValueError) as e:
-        parser = tests.any_test_help.get_parser(src)
-    assert str(
-        e.value) == 'err- This is not arf report file.'
+    with pytest.raises(Exception, match="err- This is not arf report file."):
```
matejak: Similar with the message here, I would check that either `arf` or `ARF` are present in the message string. You can use `\b` in those regexes, as they match the word boundary, so e.g. the word `arford` doesn't match.

matejak:
I have added a comment about `match` argument of `pytest.raises`, please check out all of them and consider simplifying them.

graph/xml_parser.py
```diff
@@ -61,6 +61,19 @@ def get_used_rules(self):
                         result=result.text))
         return rules
 
+    def get_notselected_rules(self):
+        ns = {
+            'ns0': 'http://checklists.nist.gov/xccdf/1.2',
```
jan-cerny: I would use `xccdf:` instead of `ns0:`.

I also suggest defining the namespace dictionary on module level and move there all the namespace definitions from the whole project, I think there is a high chance that you will want to reuse them. That can be done in a separate PR, though.

tests/test_graph.py
```diff
@@ -4,34 +4,40 @@
...
 def test_bad_tree():
-    with pytest.raises(ValueError) as e:
-        bad_tree()
-    assert str(
-        e.value) == 'err- true, false, error, unknown. noteval, notappl have not child!'
 
-    with pytest.raises(ValueError) as e:
-        tree_only_and()
-    assert str(e.value) == 'err- OR, XOR, ONE, AND have child!'
+    with pytest.raises(Exception, match="have not child!"):
```
jan-cerny: don't have any child?

tests/test_graph.py
```diff
@@ -4,34 +4,40 @@
...
+        assert tree_with_bad_value_of_operator()
+
+
+def test_bad_tree_with_bad_value_of_value():
+    with pytest.raises(Exception, match="err- unknown value"):
+        assert tree_with_bad_value_of_value()
+
+
+def test_bad_tree_with_bad_value_of_negation():
+    with pytest.raises(Exception, match="negation si bool"):
```
jan-cerny: si

tests/test_graph.py
```diff
@@ -85,7 +91,8 @@ def tree_with_bad_type():
 
 
 def tree_with_bad_value_of_negation():
-    Tree = graph.oval_graph.OvalNode(1, "value", 'true', "negovane_auto")
+    Tree = graph.oval_graph.OvalNode(1, "operator", "true", False, [
+        graph.oval_graph.OvalNode(2, "value", 'true', "negovane_auto")])
```
jan-cerny: use English

tests/test_sigmaJs.py
```diff
@@ -348,30 +348,45 @@ def test_transformation_tree_to_Json_for_SigmaJs_with_duplicated_test():
...
+    assert not parser._str_to_bool('false')
+    with pytest.raises(Exception, match="err- negation is not bool"):
+        assert parser._str_to_bool('error')
 
 
 def test_use_bat_report_file():
-    src = ('test_data/xccdf_org.ssgproject.content_rule_sssd_' +
-           'ssh_known_hosts_timeout-comment.fail.sh-xccdf_or' +
+    src = ('test_data/xccdf_org.ssgproject.content_rule_sssd_'
+           'ssh_known_hosts_timeout-comment.fail.sh-xccdf_or'
```
jan-cerny: I would split it on a more sensible place than between r and g.

matejak:
Great, PR, merging!

PR #24, opened 2019-08-28
Append new optional arguments to client

Honny1:
## New optional arguments:
- Optional arguments witch changing the choice of rules:
  - `--showFailRules` - Show only FAIL rules
  - `--showNotSelectedRules` - Show not selected rules, but you can't visualized this rules.
- Optional arguments witch changing behavior of tool:
  - `--offWebBrowser` - It does not start the web browser.
  -  `--removePassTests` - If graph have many nodes in graph, you can remove passing tests for better orientation. (Not implemented)

### DIFF
--- a/graph/client.py
+++ b/graph/client.py
@@ -40,7 +40,7 @@ class client():
                     print(rule['id_rule'] + '(Not selected)')
             return None
 
-    def get_questions(self, separator, separator1):
+    def get_questions(self, separator_rule_ids, separator_not_selected_rule_ids):
         rules = self.search_rules_id()
         if self.show_fail_rules:
             rules = self.get_only_fail_rule(rules)
@@ -48,12 +48,12 @@ class client():
             'type': 'checkbox',
             'message': 'Select rule(s)',
             'name': 'rules',
-            'choices': [separator]
+            'choices': [separator_rule_ids]
         }]
         for rule in rules:
             questions[0]['choices'].append(dict(name=rule['id_rule']))
         if self.show_not_selected_rules:
-            questions[0]['choices'].append(separator1)
+            questions[0]['choices'].append(separator_not_selected_rule_ids)
             for rule in self._get_wanted_not_selected_rules():
                 questions[0]['choices'].append(
                     dict(name=rule['id_rule'], disabled='Not selected'))
--- a/graph/client.py
+++ b/graph/client.py
@@ -28,7 +28,7 @@ class client():
             rules = self.search_rules_id()
             if self.showFailRules:
                 rules = self.get_only_fail_rule(rules)
-            for rule in rule:
+            for rule in rules:
                 print(rule['id_rule'] + r'\b')
             if self.showNotSelectedRules:
                 print('== The not selected rule IDs ==')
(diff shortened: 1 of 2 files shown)


graph/client.py
```diff
@@ -79,16 +105,37 @@ def prepare_graphs(self, rules):
...
+                webbrowser.get('firefox').open_new_tab(
+                    'html_interpreter/index.html')
+            except BaseException:
+                webbrowser.open_new_tab('html_interpreter/index.html')
 
     def parse_arguments(self, args):
         parser = argparse.ArgumentParser(
-            description='Client for visualization of SCAP rule evaluation results')
-
+            description='Client for visualization scanned rule from Security scan.')
```
jan-cerny: I think the previous description text was better.

graph/client.py
```diff
@@ -79,16 +105,37 @@ def prepare_graphs(self, rules):
...
+        parser.add_argument(
+            '--showFailRules',
```
jan-cerny: Using Camel Cases is unusual in command line options. I recommend something like `--show-fail-rules`

graph/client.py
```diff
@@ -79,16 +105,37 @@ def prepare_graphs(self, rules):
...
+            action="store_true",
+            default=False,
+            help='Show only FAIL rules')
+        parser.add_argument(
+            '--showNotSelectedRules',
+            action="store_true",
+            default=False,
+            help="Show not selected rules, but you can't visualized this rules.")
```
jan-cerny: Show notselected rules. These rules will not be visualized.

graph/client.py
```diff
@@ -34,8 +52,16 @@ def get_questions(self, separator):
         }]
         for rule in rules:
             questions[0]['choices'].append(dict(name=rule['id_rule']))
+        if self.show_not_selected_rules:
+            questions[0]['choices'].append(separator1)
```
jan-cerny: I'm confused by these separators. Why are they passed as arguments?

graph/client.py
Honny1: Yes you're right. This is a mistake after rebase.

graph/client.py
Honny1: It is for better testing. I'll rename them. Separator1 doesn't make much sense.

graph/client.py
```diff
@@ -79,16 +105,37 @@ def prepare_graphs(self, rules):
...
+        parser.add_argument(
+            '--off-web-browser',
+            action="store_true",
+            default=False,
+            help="It does not start the web browser.")
+        parser.add_argument(
+            '--remove-pass-tests',
+            action="store_true",
+            default=False,
+            help=('If graph have many nodes in graph, you can remove passing'
```
jan-cerny: This is a weird language. "Graph has a node in graph" doesn't make sense.

What about

"Remove passing tests for better orientation in large graphs."

"Remove passing tests for better orientation in graphs that contain a large amount of nodes."

Or maybe, instead of "remove"  you can use "Do not display" as they aren't removed from anywhere, they only aren't shown.

Honny1: I will use "Do not display passing tests for better orientation in graphs that contain a large amount of nodes".

PR #25, opened 2019-08-28
Adds a new tree view option.

Honny1:
## "Directory tree" view
- Created new option how to view oval tree, If tree have many nodes, you can view tree as "directory tree". 
 - Added optional arguments witch switch graph_html_interpreter to tree_html_interpreter:
   - `--tree` - Switch graph_html_interpreter to tree_html_interpreter. Rule will interpret as directory tree.


graph/client.py
```diff
@@ -89,16 +93,25 @@ def search_rules_id(self):
...
+                            sort_keys=False,
+                            indent=4) + ";"))
+                else:
+                    oval_tree = graph.oval_graph.build_nodes_form_xml(
+                        self.source_filename, rule).to_sigma_dict(0, 0)
+                    with open('graph_html_interpreter/data.js', "w+") as file:
+                        file.write("var data_json =" + str(json.dumps(
+                            oval_tree,
+                            sort_keys=False,
+                            indent=4) + ";"))
```
jan-cerny: Both if and else branch are very similar. Have you considered writing single code and parametrize it using variables?

Also, this is code is nested quite a lot. (def - try - for - if is the 4th level of indentation)  What about creating a separate function for it?

I'm not explicitly requesting changes, it's quite readable anyway. I would only suggest you thinking a few minutes about it. Feel free to disagree :smile:

graph/client.py
```diff
@@ -107,10 +120,19 @@ def prepare_graphs(self, rules):
...
         if not self.off_webbrowser:
             try:
-                webbrowser.get('firefox').open_new_tab(
-                    'html_interpreter/index.html')
+                if self.tree:
+                    webbrowser.get('firefox').open_new_tab(
+                        'tree_html_interpreter/index.html')
+                else:
+                    webbrowser.get('firefox').open_new_tab(
+                        'graph_html_interpreter/index.html')
```
jan-cerny: This seems like an unnecessary code duplication. What about creating a variable `html_path`, and calling the `webbrowser.get('firefox').open_new_tab`  method only once and passing the variable as a parameter? I think that if you will wish to replace the `webbrowser.get('firefox').open_new_tab` by anything more sophisticated it would be easier. Again, please think about it.

graph/client.py
```diff
@@ -130,12 +152,19 @@ def parse_arguments(self, args):
             action="store_true",
             default=False,
             help="It does not start the web browser.")
+        parser.add_argument(
+            '--tree',
+            action="store_true",
+            default=False,
+            help=("Switch graph_html_interpreter to tree_html_interpreter."
```
jan-cerny: I think `graph_html_interpreter` and `tree_html_interpreter` are internal details which aren't interesting for the users. Instead, I suggest something like: "Render the graph in a form of directory tree."

graph/client.py
```diff
@@ -130,12 +152,19 @@ def parse_arguments(self, args):
...
+                  " Rule will interpret as directory tree."))
         parser.add_argument(
             '--remove-pass-tests',
             action="store_true",
             default=False,
-            help=('Do not display passing tests for better orientation in'
-                  ' graphs that contain a large amount of nodes.(Not implemented)'))
+            help=(
+                'Do not display passing tests for better orientation in'
+                ' graphs that contain a large amount of nodes.(Not implemented)'))
```
jan-cerny: I have noticed that you mix single quotes and double quotes in your code. Even if they have roughly the same meaning, please try to be more consistent in using the quotes.

graph/oval_graph.py
```diff
@@ -164,6 +164,63 @@ def save_tree_to_dict(self):
...
+            "unknown": "text-dark",
+            "noteval": "text-dark",
+            "notappl": "text-dark"
+        }
+
+        VALUE_TO_ICON = {
+            "true": "glyphicon glyphicon-ok ",
+            "false": "glyphicon glyphicon-remove ",
+            "error": "glyphicon glyphicon-question-sign ",
+            "unknown": "glyphicon glyphicon-question-sign ",
```
jan-cerny: Could we avoid the trailing spaces? That's very error prone.

graph/oval_graph.py
```diff
@@ -164,6 +164,63 @@ def save_tree_to_dict(self):
...
+                "state": {
+                    "opened": True}}
+        return {
+            'text': '<span class="' + icons['color'] + '">' + self._get_label() + '</span>',
+            "icon": icons['icon'] + icons['color'],
+            "state": {
+                "opened": True},
+            'children': [
+                child.to_JsTree_dict() for child in self.children]}
+
```
jan-cerny: The 2 dictionaries differ only in the `children`  key. I suggest creating only  one dictionary, then writing a condition that optionally adds the `children` key, and then a single `return` command. That way you'll prevent code duplication again.

PR #26, opened 2019-08-28
Fix evaluation negated tests and CI Tests

Honny1:
In XML result of negated test is automatically negated.

### DIFF
--- a/graph/oval_graph.py
+++ b/graph/oval_graph.py
@@ -103,14 +103,8 @@ class OvalNode():
 
         for child in self.children:
             if child.value == 'true':
-                if child.negation:
-                    result['false_cnt'] += 1
-                else:
                     result['true_cnt'] += 1
             elif child.value == 'false':
-                if child.negation:
-                    result['true_cnt'] += 1
-                else:
                     result['false_cnt'] += 1
             elif child.value == 'error':
                 result['error_cnt'] += 1
--- a/tests/test_graph.py
+++ b/tests/test_graph.py
@@ -238,7 +238,7 @@ def test_node_value_negate():
          !f
     """
     Tree = graph.oval_graph.OvalNode(1, 'operator', 'and', False, [
-        graph.oval_graph.OvalNode(2, 'value', "false", True)
+        graph.oval_graph.OvalNode(2, 'value', "True", True)
     ]
     )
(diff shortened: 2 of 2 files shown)


graph/client.py
```diff
@@ -89,6 +89,9 @@ def parse_arguments(self, args):
         parser = argparse.ArgumentParser(
             description='Client for visualization of SCAP rule evaluation results')
 
+        parser.add_argument('--showFailRules', action="store_true", default=False, help='Show all FAIL rules')
```
jan-cerny: I think using camel cases is unusual in command line options, I guess it will confuse users. I recommend using `--show-failed-rules`. Similarly for all the remaining command line options.

jan-cerny: The help doesn't explain if it shows only failed rules (I suppose).

graph/client.py
```diff
@@ -89,6 +89,9 @@ def parse_arguments(self, args):
...
+        parser.add_argument('--showNotSelectedRules', action="store_true", default=False, help="Show not selected rules, but you can't visualized this rules.")
```
jan-cerny: .., but these rules can not be visualized.

graph/client.py
```diff
@@ -89,6 +89,9 @@ def parse_arguments(self, args):
         parser = argparse.ArgumentParser(
             description='Client for visualization of SCAP rule evaluation results')
 
+        parser.add_argument('--showFailRules', action="store_true", default=False, help='Show all FAIL rules')
+        parser.add_argument('--showNotSelectedRules', action="store_true", default=False, help="Show not selected rules, but you can't visualized this rules.")   
+        parser.add_argument('--removePassTests', action="store_true", default=False, help='If graph have many nodes in graph, you can remove passing tests for better orientation.(not implemented)')
```
jan-cerny: Too long. What about: Removes passing tests from the graph, which improves orientation in large graphs.

jan-cerny:
@Honny1 
Your PR description "In XML result of negated test is automatically negated." is confusing for anyone that wasn't present on our meeting. Instead, I would write it in the following way:

> In OVAL Results XML documents the values of `oval:definition/@result` attributes are values after the negation is applied. Therefore, we should not negate them before counting them. 

Also, if you find a statement in OVAL specification that supports this change and you quote it in PR description, you will be a superstar.

PR #27, opened 2019-09-02
XML_parser ability parse comments form oval-definition

Honny1:
Append XML_parser ability parse comments form oval-definition.

This is first PR of appending of comments form oval-definition to graphs.

### DIFF
--- a/graph/xml_parser.py
+++ b/graph/xml_parser.py
@@ -6,6 +6,13 @@ from lxml import etree as ET
 import uuid
 import graph.oval_graph
 
+global ns
+ns = {
+        'XMLSchema': 'http://oval.mitre.org/XMLSchema/oval-results-5',
+        'xccdf': 'http://checklists.nist.gov/xccdf/1.2',
+        'arf': 'http://scap.nist.gov/schema/asset-reporting-format/1.1',
+        'oval-definitions': 'http://oval.mitre.org/XMLSchema/oval-definitions-5'
+    }
 
 class xml_parser():
     def __init__(self, src):
@@ -26,11 +33,6 @@ class xml_parser():
         return result
 
     def get_data(self, href):
-        ns = {
-            'XMLSchema': 'http://oval.mitre.org/XMLSchema/oval-results-5',
-            'arf': 'http://scap.nist.gov/schema/asset-reporting-format/1.1'
-        }
-
         report_data = None
         reports = self.root.find('.//arf:reports', ns)
         for report in reports:
@@ -43,9 +45,6 @@ class xml_parser():
         return trees_data
 
     def get_used_rules(self):
-        ns = {
-            'xccdf': 'http://checklists.nist.gov/xccdf/1.2',
-        }
         rulesResults = self.root.findall(
             './/xccdf:TestResult/xccdf:rule-result', ns)
         rules = []
@@ -65,9 +64,6 @@ class xml_parser():
         return rules
 
     def get_notselected_rules(self):
-        ns = {
-            'xccdf': 'http://checklists.nist.gov/xccdf/1.2',
-        }
         rulesResults = self.root.findall(
             './/xccdf:TestResult/xccdf:rule-result', ns)
         rules = []
@@ -275,9 +271,6 @@ class xml_parser():
         return comments
 
     def prepare_definition_comments(self, oval_definitions):
-        ns = {
-            'oval-definitions': 'http://oval.mitre.org/XMLSchema/oval-definitions-5'
-        }
         definitions = []
         for definition in oval_definitions:
             comment_definition = dict(id=definition.get('id'), node=[])
@@ -309,9 +302,6 @@ class xml_parser():
         return self.help_fill_comments(comments, nodes)
 
     def insert_comments(self, data):
-        ns = {
-            'oval-definitions': 'http://oval.mitre.org/XMLSchema/oval-definitions-5'
-        }
         oval_def = self.root.findall('.//oval-definitions:definition', ns)
         comment_definitions = self.prepare_definition_comments(oval_def)
         clean_comment_definitions = []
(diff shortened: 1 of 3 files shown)


graph/xml_parser.py
```diff
@@ -204,23 +208,108 @@ def _operator_as_child(self, value, scan):
...
-    def _find_definition_by_id(self, scan, id, negate_status):
+    def _find_definition_by_id(self, scan, id, negate_status, comment):
         for definition in scan['definitions']:
             if definition['id'] == id:
                 definition['node'][0]['negate'] = negate_status
+                definition['node'][0]['comment'] = comment
                 return self._operator_as_child(definition['node'][0], scan)
+
+    def deeper_in_criteria_comments(self, criteria):
+        if criteria.get('operator') is None:
```
matejak: You can construct the comments dict as you do in the `else` branch, and if `comments['operator']` is None, make it `AND` - this saves some space.

graph/xml_parser.py
```diff
@@ -204,23 +208,108 @@ def _operator_as_child(self, value, scan):
...
+                operator='AND',
+                comment=criteria.get('comment'),
+                node=[])
+        else:
+            comments = dict(
+                operator=criteria.get('operator'),
+                comment=criteria.get('comment'),
+                node=[])
+        for criterion in criteria:
+            if criterion.get('operator') is not None:
```
matejak: Check out whether it is possible to use the `if 'operator' in criterion:` pattern instead.

graph/xml_parser.py
```diff
@@ -204,23 +208,108 @@ def _operator_as_child(self, value, scan):
...
+                comments['node'].append(
+                    self.deeper_in_criteria_comments(criterion))
+            else:
+                if criterion.get('definition_ref') is not None:
```
matejak: ditto, use the `in` operator if possible.

graph/xml_parser.py
```diff
@@ -204,23 +208,108 @@ def _operator_as_child(self, value, scan):
...
 
-    def _find_definition_by_id(self, scan, id, negate_status):
+    def _find_definition_by_id(self, scan, id, negate_status, comment):
         for definition in scan['definitions']:
             if definition['id'] == id:
                 definition['node'][0]['negate'] = negate_status
+                definition['node'][0]['comment'] = comment
                 return self._operator_as_child(definition['node'][0], scan)
+
+    def deeper_in_criteria_comments(self, criteria):
```
matejak: The method name is not descriptive. What should it return? It looks like that given criteria, it recursively constructs a tree containing operators and comments.

graph/xml_parser.py
```diff
@@ -204,23 +208,108 @@ def _operator_as_child(self, value, scan):
...
+                else:
+                    comments['node'].append(
+                        dict(
+                            value_id=criterion.get('test_ref'),
+                            comment=criterion.get('comment')))
+        return comments
+
+    def prepare_definition_comments(self, oval_definitions):
+        ns = {
+            'oval-definitions': 'http://oval.mitre.org/XMLSchema/oval-definitions-5'
```
matejak: This looks like a constant that should be together with other namespaces, unless there is a reason not to do it.

graph/xml_parser.py
```diff
@@ -204,23 +208,108 @@ def _operator_as_child(self, value, scan):
...
+        for definition in array:
+            if definition_['id'] == definition['id']:
+                return True
+        return False
+
+    def help_fill_comments(self, comments, nodes):
+        out = nodes
+        if 'operator' in out:
+            out['comment'] = comments['comment']
+            for i in range(len(out['node'])):
```
matejak: The loop contains occurences of `out['node'][i]` and `comments['node'][i]`, check out how to use `enumerate` and `zip` in `for` loops to make this more Pythonic with less square brackets.

graph/xml_parser.py
```diff
@@ -204,23 +208,108 @@ def _operator_as_child(self, value, scan):
...
+        return [out]
+
+    def fill_comment(self, comment_definition, data_definition):
+        comments = comment_definition['node'][0]
+        nodes = data_definition['node'][0]
+        return self.help_fill_comments(comments, nodes)
+
+    def insert_comments(self, data):
+        ns = {
+            'oval-definitions': 'http://oval.mitre.org/XMLSchema/oval-definitions-5'
```
matejak: ditto, we have a repeated constant here.

graph/xml_parser.py
```diff
@@ -204,23 +208,108 @@ def _operator_as_child(self, value, scan):
...
+            definitions.append(comment_definition)
+        return definitions
+
+    def is_definition_in_array(self, definition_, array):
+        for definition in array:
+            if definition_['id'] == definition['id']:
+                return True
+        return False
+
+    def help_fill_comments(self, comments, nodes):
```
matejak: This method is also recursive. I recommend to reflect the recursive nature in the method's name.

graph/xml_parser.py
```diff
@@ -6,6 +6,14 @@
 import uuid
 import graph.oval_graph
 
+global ns
```
matejak: Global here makes no difference. You use `global` e.g. inside functions when you want to modify the global object. Read-only access works without `global`.

graph/xml_parser.py
```diff
@@ -59,13 +59,12 @@ def get_used_rules(self):
                         id_rule=ruleResult.get('idref'),
                         id_def=check_content_ref.attrib.get('name'),
                         href=check_content_ref.attrib.get('href'),
-                        result=result.text))
+                        result=result.text
```
matejak: Also add a trailing line, i.e. `result=result.text,`
That way, adding another member of the dictionary will be just about adding one more line.
Concerning those brackets, it is OK to close both of them at once to save vertical space.

matejak:
Looks great, merging!

PR #28, opened 2019-09-03
Implementation comments to OvalNode

Honny1:
This PR implement `comment` parameter to `OvalNode` and functions, which work with comments and display comments in graph. 

This is second PR of appending of comments form oval-definition to graphs.


graph/oval_graph.py
```diff
@@ -199,10 +203,17 @@ def _get_node_icon(self):
             color=VALUE_TO_COLOR[value],
             icon=VALUE_TO_ICON[icon])
 
+    def get_comment(self):
+        if self.comment is not None:
+            return '<i>' + str(self.comment) + '</i>'
```
matejak: You have data formatting as part of the `OvalNode`, but you probably want to have it in another layer, so when you change your mind and you want to format the output in a different way, you don't have to modify the core code. In other words, think of how to separate data extraction from the data "styling".

Honny1:
@matejak Bug with displaying comments fixing #31 PR.

matejak:
This PR is another step to full display of comments in the output. Another PRs fix issues that still remain. The `OvalNode` refactor proposed by me earlier is out of scope of this PR, so I am merging it. Thank you for your work!

PR #34, opened 2019-09-10
Refactor oval node

Honny1:
This PR refactors `ovalNode`.

- It removes methods from `ovalNode`, that can convert the internal tree data structure to JSON and moves them to a new `converter` class.
- It refactors the `ovalNode` constructor, adds new methods for input validation.


graph/converter.py
```diff
@@ -0,0 +1,357 @@
...
+            icon, value = self.tree.value, icon
+        else:
+            if self._is_negated_boolean('true', value):
+                icon = 'false'
+            elif self._is_negated_boolean('false', value):
+                icon = 'true'
+            else:
+                icon = value
+
+        VALUE_TO_COLOR = {
```
matejak: If dictionaries like this are renamed and moved to class top level, it will make style overrides even easier.

graph/converter.py
```diff
@@ -0,0 +1,357 @@
...
+        return positions
+
+    def _convert_nodes_in_rows_to_nodes(self, nodes_in_rows):
+        nodes = []
+        for row in nodes_in_rows:
+            for node in nodes_in_rows[row]:
+                nodes.append(node)
+        return nodes
+
+    def _change_position(self, nodes_in_rows):
```
matejak: Code in this function is quite complex. It may not be worth to change, but let's keep an eye on code like this in the future.

graph/converter.py
```diff
@@ -0,0 +1,350 @@
...
+            color=self.VALUE_TO_BOOTSTRAP_COLOR[values['negation_color']],
+            icon=self.VALUE_TO_ICON[values['out_color']],
+        )
+
+    def get_comment(self):
+        if self.tree.comment is not None:
+            return str(self.tree.comment)
+        return ""
+
+    def to_JsTree_dict(self):
```
matejak: Make a JSTree converter class that would contain a `to_dict` method, and a Sigma converter class with another `to_dict` method. It would be best to have a base `Converter` class, and those `JSTree` and `Sigma` would derive from it.

matejak:
I like your changes, but there is still something left to do.
The current `converter` class can do both, which is confusing. When one is interested only in the `JSTree` output, one is confused by the class code which is used relevant for the Sigma output. So please split the class into two classes, or one base class and two derived classes.

graph/converter.py
Honny1: I plan to remove `SigmaJs` from the project. It is therefore not necessary to create inheritance classes.

matejak:
> I plan to remove SigmaJs from the project. It is therefore not necessary to create inheritance classes.

OK, fair enough.

PR #35, opened 2019-09-10
Changed label of the nodes 

Honny1:
This PR changes the labels of the nodes. For better understanding graph
- If operator node is negated. It changes AND and OR to &not; AND and &not; OR. 
- It uses uppercase letters for operator labels.
- if test node is negated. It changes `label of the node` to &not; `label of the node` . 
- Changed coloring of the nodes if is negated. 
- In JsTree changed icons of the node if is negated


tests/test_data/referenc_result_data_graph.js
```diff
@@ -42,13 +42,21 @@ var data_json ={
     ],
     "edges": [
         {
+<<<<<<< HEAD
             "id": "UUID-HASH",
+=======
+            "id": "e4fd7f98-fd2d-44ef-ba32-181b23af0ba3",
+>>>>>>> 653e5a9... Changed label of negated operator and update test files
```
matejak: Conflict!

matejak:
Test data have a conflict. What is really strange is how those tests could pass when the JSON is invalid due to conflict markers.

Honny1:
Is not it strange. The test reads the file as a text file and then creates an array of lines in the file without `edges`. Because `edges` contain `UUID` data. That's why the tests pass. The file is `.js` and it starts with `var data =" json..."` because it bypasses the request [`CORS is not HTTP`](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS/Errors/CORSRequestNotHttp). In fact, it is not a `JSON` file and the test compares the lines in the file.

tests/test_sigmaJs.py
```diff
@@ -128,7 +128,7 @@ def test_create_node_with_negation_dict_for_sigmaJs():
         'borderColor': '#ff0000',
         'color': '#00ff00',
         'id': 1,
-        'label': 'and',
+        'label': 'NAND',
```
jan-cerny: I don't like using NAND or NOR in the OVAL visualization. OVAL specification doesn't define NAND and NOR. OVAL doesn't use "normal" binary logic, it defines the operations over different set. Saying that, people can be confused by meaning of "NAND" or "NOR" here. Also, they will not find these operations in their OVAL definitions.

I'm not saying it's bad, but I'm not sold on that either. Could you please elaborate on why you have chosen this solution? I'm missing any rationale in the PR description.

Have you considered having a special node for the negate operation? An OVAL criterion which involves negation would be represented by 2 nodes in the graph - child node containing the OVAL operation (AND, OR, etc.) and parent node containing the negation?

Honny1: Good point. I used binary logic to display negation. And what about using an exclamation mark to show negation and create some description? I think creating a special node is a good idea, but at the moment it will be need big change. In the future it will be good to create. You can creat issue.

Honny1: Or I can use propositional logic. And use &not; `&not;` to display the negation. I think it will be better. What do you think? @jan-cerny

Issue #38, opened 2019-09-16
Output html directory 


Honny1:
Change the default graph output directory. The current solution is not good, because it is inside the package. And the user needs permission to run the command.

PR #42, opened 2019-09-16
Output of html

Honny1:
This PR changes the output of the graph to the directory in /tmp.
Fixes: #38

### DIFF
--- a/oval_graph/client.py
+++ b/oval_graph/client.py
@@ -144,20 +144,16 @@ class client():
         except Exception as error:
             raise ValueError('Rule: "{}" Error: "{}"'.format(rule, error))
 
-    def open_web_browser(self, rule, date):
+    def open_web_browser(self, name_dir):
         if not self.off_webbrowser:
+            src = os.path.join(
+                tempfile.gettempdir(),
+                'graph-of-' + name_dir,
+                'index.html')
             try:
-                webbrowser.get('firefox').open_new_tab(
-                    os.path.join(
-                        tempfile.gettempdir(),
-                        'graph-of-' + rule + date,
-                        'index.html'))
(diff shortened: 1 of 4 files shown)


oval_graph/client.py
```diff
@@ -105,29 +109,54 @@ def create_dict_of_rule(self, rule_id):
...
             raise ValueError('Rule: "{}" Error: "{}"'.format(rule, error))
 
-    def open_web_browser(self):
+    def open_web_browser(self, rule, date):
         if not self.off_webbrowser:
             try:
-                webbrowser.get('firefox').open_new_tab(self.xml_parser.get_src(
-                    self.html_interpreter + '/index.html'))
+                webbrowser.get('firefox').open_new_tab(
+                    os.path.join(
```
matejak: The path looks the same, why not to save it to a variable and use it here and in the except block?

oval_graph/client.py
```diff
@@ -105,29 +109,54 @@ def create_dict_of_rule(self, rule_id):
             return converter.to_JsTree_dict()
         return converter.to_sigma_dict(0, 0)
 
-    def save_dict(self, dict):
-        with open(self.xml_parser.get_src(self.html_interpreter + '/data.js'), "w+") as data_file:
+    def save_dict(self, dict_, rule, date):
+        with open(os.path.join(tempfile.gettempdir(),
+                               'graph-of-' + rule +
```
matejak: This dirname construction repeats in several places. Why not to have a function that will create the `empfile.gettempdir() / 'graph-of-' + rule + date` directory when you supply `rule` and `date`? Also check the `pathlib` module, it enables you to construct paths using the / operator without having to call `os.path.join()` function.

matejak:
Thank you for improving the PR, merging!

PR #45, opened 2019-09-17
Remove sigma js

Honny1:
This PR removes the option to display an oval tree as a graph because the graph is sometimes unclear.


matejak:
RIP, I have liked the full-blown 2D graph, but the new expandable tree is sleek, and maintaining the two backends at the same time is not feasible.

PR #46, opened 2019-09-17
Change imports

Honny1:
This PR changes:
 - Imports in project 
 - Class name ovalNode to oval_node
 - File name oval_graph.py to oval_node.py

### DIFF
--- a/oval_graph/client.py
+++ b/oval_graph/client.py
@@ -1,15 +1,15 @@
-from __future__ import print_function, unicode_literals
-from datetime import datetime
 import re
-import oval_graph.xml_parser
-import oval_graph.oval_graph
-import oval_graph.converter
 import webbrowser
 import json
 import argparse
 import tempfile
 import os
 import shutil
+from datetime import datetime
+
+from .xml_parser import xml_parser
+from .oval_node import oval_node
+from .converter import converter
 
 
 class client():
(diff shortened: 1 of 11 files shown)


oval_graph/client.py
```diff
@@ -1,16 +1,20 @@
 from __future__ import print_function, unicode_literals
```
matejak: I don't think that we need this import, as the project is python3-only.

oval_graph/oval_node.py
```diff
@@ -1,18 +1,18 @@
...
+import oval_graph.evaluate
+
 
 '''
     This module contains methods and classes for
     constructing and controlling an oval tree.
 '''
 
 
-class OvalNode():
```
matejak: Why that? The [standard recommends CapWords](https://www.python.org/dev/peps/pep-0008/#class-names)

oval_graph/command_line.py
```diff
@@ -9,18 +9,18 @@ def print_where_is_saved_result(results_src):
 
 
 def main():
-    client = oval_graph.client.client(sys.argv[1:])
-    rules = client.search_rules_id()
+    client_ = client(sys.argv[1:])
```
matejak: I don't like the variable name `client_`. What about making the current class name `client` uppercase `Client` according to the standard? The, you could have the variable `client` as before.

oval_graph/client.py
```diff
@@ -1,16 +1,20 @@
...
-import oval_graph.oval_graph
-import oval_graph.converter
 import webbrowser
 import json
 import argparse
 import tempfile
 import os
 import shutil
 
+from datetime import datetime
```
matejak: `datetime` is a Python standard library import, so no need to put it aside from other std imports.

matejak:
Looks much better, works great. Thank you for the PR!

PR #48, opened 2019-10-01
Removed py library usage

Honny1:
This PR remove `py` library, because usage in packages is problematic.


matejak:
LGTM, good riddance!

PR #49, opened 2019-10-01
Changed PyInquirer library to inquirer

Honny1:
This PR changes PyInquirer library to inquirer. Because PyInquirer has problems with installation.


Honny1:
@matejak because inquirer library can't disable select of  not selected rules.

Honny1:
@matejak  OK I understand. I have an idea, how to keep the option in both cases.

matejak:
Works great, merging!

PR #51, opened 2019-10-24
Prepare on build

Honny1:
These PRs include:
- deleted forgotten include
- changed wording client to tool

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -13,8 +13,8 @@ def get_long_description():
 
 
 setup(name='oval_graph',
-      version='0.0.5',
-      description='Client for visualization of SCAP rule evaluation results',
+      version='0.0.1',
+      description='Tool for visualization of SCAP rule evaluation results',
       long_description=get_long_description(),
       long_description_content_type="text/markdown",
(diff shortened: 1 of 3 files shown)


matejak:
It is not a good practice to include the `.spec` file in the repository, because it changes independently from the package. For example, package maintainer may decide to rebuild it, which changes the release number, and you may have even more than one spec file - for example for the new and former Fedora versions. Of course, you host the .spec file in a git repository, but not in the Github with the project.
The other changes are good, there is no problem in merging them.

Honny1:
@matejak  Well, I'll remove the `.spec` file.

setup.py
```diff
@@ -13,8 +13,8 @@ def get_long_description():
 
 
 setup(name='oval_graph',
-      version='0.0.5',
-      description='Client for visualization of SCAP rule evaluation results',
+      version='0.0.1',
```
matejak: We have to do something about this, but that's for another PR.

matejak:
Tests were failing but they work again, so merging, the versioning problem can be addressed later.

PR #53, opened 2019-10-31
Save the output files to the selected directory 

Honny1:
This PR include:

-  a new `--out` option that allows you to choose where to save output files 
- rework test
 
Fixes: #53

### DIFF
--- a/oval_graph/client.py
+++ b/oval_graph/client.py
@@ -175,7 +175,7 @@ class Client():
             '--out',
             action="store",
             default=False,
-            help="Save the output files where it is defined.")
+            help="The directory where to save output files.")
         parser.add_argument(
             '--remove-pass-tests',
             action="store_true",
--- a/oval_graph/client.py
+++ b/oval_graph/client.py
@@ -21,6 +21,7 @@ class Client():
         self.off_webbrowser = self.arg.off_web_browser
         self.source_filename = self.arg.source_filename
         self.rule_name = self.arg.rule_id
+        self.out = self.arg.out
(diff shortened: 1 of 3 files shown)


oval_graph/client.py
```diff
@@ -162,6 +171,11 @@ def parse_arguments(self, args):
             action="store_true",
             default=False,
             help="It does not start the web browser.")
+        parser.add_argument(
+            '--out',
+            action="store",
+            default=False,
+            help="Save the output files where it is defined.")
```
matejak: You probably mean `The directory where to save output files`.

oval_graph/client.py
```diff
@@ -162,6 +171,11 @@ def parse_arguments(self, args):
             action="store_true",
             default=False,
             help="It does not start the web browser.")
+        parser.add_argument(
+            '--out',
+            action="store",
+            default=False,
```
matejak: If you don't specify the default, it will be `None`, which is probably good enough.

oval_graph/client.py
```diff
@@ -118,14 +119,22 @@ def copy_interpreter(self, dst):
             else:
                 shutil.copy2(s, d)
 
+    def get_save_src(self):
+        if isinstance(self.out, str):
```
matejak: `if self.out` is good enough, although `isinstance` is not wrong. The simpler way to check whether `self.out` is a non-empty string, i.e. "truthy", is more Pythonic.

matejak:
Looks great, thanks!

Issue #39, opened 2019-09-16
Save and load graph/report as JSON


Honny1:
Will be good, if user will have option save to small JSON of graph/report. And restore them.

PR #57, opened 2019-11-05
Add command arf-to-json

Honny1:
This PR include new command `arf-to-json` which transform rule form arf to json.

Fixes: #39 
Restore to graph command will be append in next PR.

### DIFF
--- a/oval_graph/command_line.py
+++ b/oval_graph/command_line.py
@@ -1,6 +1,7 @@
 import sys
 
 from .arf_to_html import ArfToHtml
+from .arf_to_json import ArfToJson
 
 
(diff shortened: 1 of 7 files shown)


oval_graph/command_line.py
```diff
@@ -23,4 +37,4 @@ def arf_to_html():
 
 
 if __name__ == '__main__':
-    arf_to_html()
+    globals()[sys.argv[1]](sys.argv[2:])
```
matejak: I would remove this completely - `globals()[sys.argv[1]](sys.argv[2:])` is difficult to maintain and difficult to understand.

matejak:
LGTM, thanks!

PR #63, opened 2019-12-06
Append offline sources for bootstrap, jquery, etc. in report

Honny1:
This PR changes sources for bootstrap, jquery, etc. in report. If user not able to connect to internet.


matejak:
Great PR, removes lots of lines, while the result can be viewed offline without any issues. Merging!

PR #64, opened 2019-12-13
Json output structure

Honny1:
This PR changes json output structure to store more than one rule in json.


matejak:
How should it work? If I specify a regex that matches multiple rules in ARF, it tells me to narrow it down to one rule only.

Honny1:
This is fine if the inquirer package is not installed. If the inquirer package is installed, it displays a selection of rules, then the user can select multiple rules. I plan to add an option with a selection of all the matched rules.

matejak:
This works great, you choose to use a dictionary to store those graph representations, which is smart.
I recommend to save JSON output to one file - right now, the `--output XXX` creates a directory XXX with multiple files, while this PR allows for saving everything to a single file.

Issue #62, opened 2019-12-05
 Redirecting command arf-to-json to file


Honny1:
Problem with redirecting command arf-to-json to file when rule name is not unique.

matejak:
Please tell us what exactly is the problem. Also, does it have connection to the Inquirer?

Honny1:
Yes, there is a problem with the Inquirer. When part of the rule name or regex is used as the rule name. A selection of rules appears, but when the command is redirected to a file. Everything that is printed on the command line is saved to a file. Does not show any selection.

PR #65, opened 2019-12-13
Fixed redirecting output of command to file

Honny1:
This PR contain:
- fix of bug #62 
- Added quotation mark to rules names in selection without installed inquirer for easter copy and paste 

Fixes: #62

### DIFF
--- a/oval_graph/arf_to_json.py
+++ b/oval_graph/arf_to_json.py
@@ -4,13 +4,25 @@ import os
 import shutil
 import pprint
 from datetime import datetime
-
+import sys
 
 from .converter import Converter
 from .client import Client
 
 
 class ArfToJson(Client):
+
+    def run_gui_and_return_answers(self):
+        try:
+            if sys.stdout.isatty():
+                import inquirer
+                return inquirer.prompt(self.get_questions())
+            else:
+                return {'rules': [
+                    rule['id_rule'] for rule in self.search_rules_id()]}
+        except ImportError:
+            print(self.get_selection_rules())
+            return None
 
     def create_dict_of_rule(self, rule_id):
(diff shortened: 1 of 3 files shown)


oval_graph/client.py
```diff
@@ -31,20 +31,24 @@ def run_gui_and_return_answers(self):
...
-                print('== The not selected rule IDs ==')
-                for rule in self._get_wanted_not_selected_rules():
-                    print(rule['id_rule'] + '(Not selected)')
-            print("You haven't got installed inquirer lib. "
-                  "Please copy id rule with you want use and put it in command")
+            print(self.get_selection_rules())
             return None
 
+    def get_selection_rules(self):
+        out = '== The Rule IDs ==\n'
```
matejak: When you create a long multi-line string, it is best to create a list of lines, and make the string from list of lines using `"\n".join(lol)` as the last step.

This method also mixes data presentation and data gathering, I recommend to split it - first gather list of rules, and transform the list to a string that should be displayed in another method.

oval_graph/arf_to_json.py
```diff
@@ -4,14 +4,26 @@
...
-
+import sys
 
 from .converter import Converter
 from .client import Client
 
 
 class ArfToJson(Client):
 
+    def run_gui_and_return_answers(self):
```
matejak: It would be good to have a possibility to disable the `inquirer` even if it is installed. Utilities s.a. `ls` or `grep` also allow to switch off color output, although they can detect whether it is a good idea to show colors or not.

Honny1: I plan to add an option to deactivate the selection that will process all the matching rules.

matejak:
I see, when output is redirected to a file, the Inquirer selection is skipped, and all matching rules are considered as selected.
Works great, merging!

PR #69, opened 2019-12-21
New option for process all matched rules

Honny1:
This PR include option ```--all``` witch process all matched rules.

### DIFF
--- a/oval_graph/arf_to_html.py
+++ b/oval_graph/arf_to_html.py
@@ -8,7 +8,6 @@ class ArfToHtml(Client):
     def __init__(self, args):
         super().__init__(args)
         self.off_webbrowser = self.arg.off_web_browser
-        self.all_rules = self.arg.all
 
     def prepare_data(self, rules):
         try:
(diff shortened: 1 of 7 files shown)


oval_graph/arf_to_html.py
```diff
@@ -12,6 +12,7 @@ def __init__(self, args):
     def prepare_data(self, rules):
         try:
             out = []
+            rule = None
```
matejak: This shouldn't be needed - `rule` is overwritten in the for loop.

oval_graph/json_to_html.py
```diff
@@ -38,10 +40,33 @@ def load_rule_names(self):
...
+        questions = [
+            checkbox(
+                'rules',
+                message=(
+                    "= The Rules IDs = (move - UP and DOWN arrows,"
+                    " select - SPACE or LEFT and RIGHT arrows, submit - ENTER)"),
+                choices=choices_,
+            ),
+        ]
+        return questions
```
matejak: I think that this is code that is duplicated in `client.py`.

matejak:
Nice adjustments, thanks!

Issue #67, opened 2019-12-19
Output to single JSON file.


matejak:
The when `--output XXX` argument is given to arf-to-json, it saves JSONs into a directory XXX, which contains one or more files. It would be better not to create any directories, and to save everything into a file XXX.

PR #70, opened 2019-12-23
Output to single JSON file

Honny1:
This PR change function of parameter ```--out``` in command ```arf-to-json```.  User can store rules in one json file. 
Fixes: #67

### DIFF
--- a/oval_graph/client.py
+++ b/oval_graph/client.py
@@ -176,8 +176,16 @@ class Client():
 
     def parse_arguments(self, args):
         self.prepare_parser()
+        self.prepare_parser_out()
         args = self.parser.parse_args(args)
         return args
+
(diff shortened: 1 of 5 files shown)


oval_graph/client.py
```diff
@@ -176,9 +176,17 @@ def get_save_src(self, rule):
 
     def parse_arguments(self, args):
         self.prepare_parser()
+        self.prepare_parser_out()
         args = self.parser.parse_args(args)
         return args
 
+    def prepare_parser_out(self):
+        self.parser.add_argument(
+            '--output',
```
matejak: Just a small suggestion - add here also `-o`, so one can specify output with minimum amount of writing.

matejak:
Awesome, thanks!

PR #71, opened 2019-12-30
New parameter in json-to-graph

Honny1:
These PR contains:
- A new parameter `rule_name` in command `json-to-graph`. That can be used to search a rule stored in a json file.
- rework the json-to-graph class
- rework arg parser in Client class


matejak:
This PR adds the rule selection interface to the `json-to-html`, and it works pretty good.
Great work!

Issue #21, opened 2019-08-19
Distinguish between OVAL tests and OVAL definitons


jan-cerny:
It would be nice if we could spot if a graph node is an OVAL test, OVAL definition or a criteria. Think of a text in label, different color or shape of the node. That would be helpful when a defintion contains another definition reference using extend_definition.

PR #74, opened 2020-01-09
Labels for OVAL tests and OVAL definitions

Honny1:
This PR contains:
- labels with information about node type. (Rule/Definition/Extend definition/Criteria/Test)
- Description of oval-def in second node
 
Fixes: #21 #43

### DIFF
--- a/oval_graph/oval_node.py
+++ b/oval_graph/oval_node.py
@@ -22,6 +22,7 @@ class OvalNode():
         input_value (str): value of node
         input_negation (bool): value indicating whether the node is negated or not
         comment (str): text about node
+        tag (str): tag specifies if the node represents OVAL test, OVAL definition or XCCDF rule
         children ([OvalNode]): array of children of node
 
     Attributes:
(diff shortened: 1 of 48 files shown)


oval_graph/oval_node.py
```diff
@@ -22,6 +22,7 @@ class OvalNode():
         input_value (str): value of node
         input_negation (bool): value indicating whether the node is negated or not
         comment (str): text about node
+        tag (str): info about node if is test node, definition node or rule node
```
jan-cerny: You probably mean that tag specifies if the node represents OVAL test, OVAL definition or XCCDF rule.

That leads me to this question: Would it be better to use some enumeration for this instead of arbitrary string?

Honny1:
@jan-cerny I mean put description form oval definition to node with tag `Definition`. It's usually node under node with name rule.

oval_graph/oval_node.py
Honny1: Enumeration sounds good. I started think about implementation enumeration for values and operators too. I will implement enumeration in following PR.

jan-cerny:
This is a huge improvement. Thank you!

PR #75, opened 2020-01-14
Build script

Honny1:
This PR contains build script for automated releases.


build.sh
```diff
@@ -0,0 +1,51 @@
...
+vim ${module}/version.py
+vim ~/rpmbuild/SPECS/${name}.spec
+
+version=$(python3 setup.py --version)
+
+# Version commit
+git add -A
+git commit -m "${version}"
+git tag "${version}"
+git push upstream master
```
matejak: In order to push a tag, you need `git push --follow-tags`

build.sh
```diff
@@ -0,0 +1,51 @@
...
+
+# Edit version, change log, etc... 
+vim ${module}/version.py
+vim ~/rpmbuild/SPECS/${name}.spec
+
+version=$(python3 setup.py --version)
+
+# Version commit
+git add -A
+git commit -m "${version}"
```
matejak: You don't want to do `git add -A` in a script - it can pick up any trash that is present in the project dir. I would say don't even commit in the script, do the last commit manually, and when everything is OK, just tag.

build.sh
```diff
@@ -0,0 +1,51 @@
+name="oval-graph"
+module="oval_graph"
+
+git checkout master
+
+# Edit version, change log, etc... 
+vim ${module}/version.py
```
matejak: Use `sed` to substitute version numbers automatically. For that, you need version.
Something like `sed -i /version=.*/version=$version/ file_with_version` could do the trick.

build.sh
```diff
@@ -0,0 +1,51 @@
...
+version=$(python3 setup.py --version)
+
+# Version commit
+git add -A
+git commit -m "${version}"
+git tag "${version}"
+git push upstream master
+
+# Build for Pipy
+rm -r dist
```
matejak: `rm -rf` is better - that way, it will be fully interactive.

build.sh
```diff
@@ -0,0 +1,51 @@
...
+python3 setup.py pytest sdist bdist_wheel
+twine upload dist/*
+
```
matejak: I would end the release script here, and add a build script for Fedora to some subdirectory, as it is not directly related to the project.

build.sh
```diff
@@ -0,0 +1,51 @@
+name="oval-graph"
```
matejak: It is a good idea to have `set -e` at the beginning, so when there is an error and any command fails, the script aborts.

build.sh
```diff
@@ -0,0 +1,33 @@
...
+module="oval_graph"
+
+if [ "$1" != "" ]; then
+    new_version=$1
+    if [[ $new_version =~ ^[0-9]{1,}.[0-9]{1,}.[0-9]{1,}$ ]]; then
+        old_version=$(python3 setup.py --version)
+        git checkout master
+        git pull upstream master
+        changes=$(git status --porcelain=v1 | grep -q '^\(.M\|M.\)')
+        if [$changes == ""];then
```
matejak: This won't work - when grep is executed with `-q`, there is no output even if the repository is not clean. You need to check for the return code of `grep` directly, for example by putting as the condition statement to that `if`. Just keep in mind that if `grep` returns success, it means that there were changes.

setup.py
```diff
@@ -1,6 +1,7 @@
 import sys
 from setuptools import setup, find_packages
 from setuptools.command.test import test as TestCommand
+from oval_graph.version import __version__
```
matejak: This will import your installed `oval_graph`'s version attribute. You probably need to add the `setup.py`'s directory to the `sys.path`. Check out [another setup.py](https://github.com/matejak/imreg_dft/blob/master/setup.py) for inspiration.

matejak:
Changes are directionally good, but there are still some more adjustments needed.

matejak:
Looks fine - good luck with upcoming releases!

PR #76, opened 2020-01-14
Hide pass tests in report

### LINKED ISSUE #22: Option to show only the failing tests
Does it make sense to have an option to show only the failing tests? As you can see, sometimes is hard to navigate between the tests
Untitled

Honny1:
This PR contains:
- new parameter `--hide-passing-tests`  
This parameter hide all pass tests for better orientation in graph. (see pictures)
### Not used  `--hide-passing-tests`
alt text

### Used  `--hide-passing-tests`
alt text
Fixes #22


oval_graph/converter.py
```diff
@@ -29,6 +30,17 @@ def __init__(self, tree):
...
+        self.HIDE_PASS_TESTS = {
+            True: {  # Key is value of hide_pass_tests option
+                "true": False,  # Key is result of node
+                "false": True
+            },
+            False: {
+                "true": True,
+                "false": True
+            }
+        }
```
jan-cerny: This is really interesting piece of code. I think that you can achieve your goal without this mind-boggling dictionary.

oval_graph/converter.py
```diff
@@ -51,7 +63,7 @@ def get_tag(self):
             return str(self.tree.tag)
         return ""
 
-    def to_JsTree_dict(self):
+    def to_JsTree_dict(self, option=False):
```
jan-cerny: I think `option` isn't a good variable name, because it doesn't specify which option, what the option does. Try to name your variables in a self-explanatory way. For example `hide_passing_tests` could be a good name, as it's obvious what happens when you set the variable to true.

oval_graph/converter.py
```diff
@@ -65,12 +77,20 @@ def to_JsTree_dict(self):
...
                                for child in self.tree.children]
         return out
 
+    def _hide_node(self, option):
+        value = self.tree.evaluate_tree()
+        if value is None:
+            value = self.tree.value
+        if value in self.HIDE_PASS_TESTS[option].keys():
+            return self.HIDE_PASS_TESTS[option][value]
+        return True
```
jan-cerny: I'm not sure if I understand it. Correct me if I'm wrong. Probably, it means that:
* if user doesn't provide `--hide-pass-tests` it always returns True.
* if user provides  `--hide-pass-tests`:
  * if the node is "true" returns False
  * otherwise returns True

I think it can be rewritten to make the code more readable in this manner, and you won't need the dictionary.

Also, the function name `_hide_node` is confusing. Based on the name I would expect that when it returns True it means that the node will be hidden, but it seems it's the opposite - when it returns False the node will be hidden. So I recommend either swapping True and False return values or renaming the function to `_show_node`.

oval_graph/arf_to_html.py
```diff
@@ -31,3 +37,10 @@ def prepare_parser(self):
             action="store_true",
             default=False,
             help="It does not start the web browser.")
+        self.parser.add_argument(
+            '--hide-pass-tests',
```
jan-cerny: `--hide-passing-tests`?

jan-cerny:
Also, a good practice is to fill in the description in the pull requests. That helps the reviewers to understand the change you propose. I think that in this case a screen shot would be great!

PR #89, opened 2020-01-30
HTML report as one file

Honny1:
This PR contains:
- outputs of commands `art-to-graph` and `json-to-graph` are one html file.


jan-cerny:
The code changes look fine to me.

I doubt that having hardocded fonts and Javascripts in the test data is useful. Do you think that it could be sufficient in our tests to just search for patterns in the generated HTML files? That would save your work because you will not have to change the tests each time you modify something in the HTML generator. What is your opinion?

Honny1:
It is good idea. @jan-cerny

Issue #66, opened 2019-12-19
Embed OVAL details into the graph


matejak:
The oscap HTML report with OVAL details (default in oscap>=1.3) contains list of tests, and under each test, one can usually see detailed information, usually collected objects and a hint that either objects were found, but not expected, or that they were expected, but weren't found etc.
It would be nice if one could click on a test result, and the test details would pop up, or to show test details by another means.

jan-cerny:
@matejak @Honny1 What about including an "advertisment" on the tool into the HTML report? Probably only on Fedora at this moment, but it would be nice.

Honny1:
@jan-cerny I think it is good idea.

Honny1:
@jan-cerny @matejak I have an idea. And what to create a command that put the charts into an html report. Command should have parameters arf and html report. Is it good idea?

PR #91, opened 2020-02-12
Embeds OVAL details about test to graph

Honny1:
This PR embeds OVAL details about test to graph. After click on some test display OVAL details about test. Does not implement reference variables.

Fixes: #66


oval_graph/parts/footer.html
```diff
@@ -1,13 +1,137 @@
...
+							break;
+						case 'exec':
+							out += "x";
+							break;
+						default:
+							// pass
+							break;
+					}
+				} else {
+					out += "-";
```
matejak: This is not so simple - in OVAL, we may either say that the state is `uread` on or off, which you handle well. But if we don't care, you will visualize it the same as if we want it off. I suggest using e.g. `?` to indicate the "we don't care" option.

oval_graph/parts/footer.html
```diff
@@ -1,13 +1,137 @@
...
+			var index_special_char = str.indexOf('@');
+			var text = str.substring(0, index_special_char != -1 ? index_special_char : str.length);
+			var text_with_spaces = text.replace("_", " ");
+			return text_with_spaces.charAt(0).toUpperCase() + text_with_spaces.slice(1);
+		}
+
+		function create_table(type, table_object) {
+			switch (type) {
+				case 'file_object':
+					table_object = prepare_file_object(table_object);
```
matejak: Try to check out if you can detect whether a function `prepare_<type>` exists, and if yes, apply it on the table_object. That way, you will get rid of the `switch` and you make addition of "js visualization plugins" easier.
If it's too complicated, let's do it in another PR.

tests/test_graph.py
```diff
@@ -52,78 +52,78 @@ def bad_tree():
          t
     """
     t = OvalNode(
-        1, "value", "true", False, None, None,  [
+        1, "value", "true", False, None, None, None, [
```
matejak: General remark: You seem to have to update tests quite a lot if you make a change in API. In cases like this, it would be worth thinking to use named arguments and arguments with default values, so you don't have to alter all those function invocations except cases when you want to pass something else than `None`.
Named arguments help against confusion what has been passed and what is supplied by default.

Honny1: I'll do it in another PR.

matejak:
There is an issue with a file permission-related rule `file_permissions_library_dirs`:

- The OVAL specifies (searches for) objects that have `gwrite` and `owrite` set.
- The summary indicates that it specifies objects with all permissions unset
image

matejak:
Great enhancement! Some issues may pop up, as introduction of OVAL objects is a huge thing, but they can be addressed in subsequent PRs.

PR #94, opened 2020-02-25
Guide

Honny1:
This PR contains:
- guide for user
- fix typos in README.md
- update pictures

PR #95, opened 2020-02-27
Rework OvalNode

### LINKED ISSUE #37: Rework OvalNode
- Use named arguments in constructor of OvalNode 
- Improve the code style in help function for the OVAL operators evaluation
- Reduce the susceptibility of OvalNode changes

Honny1:
This PR contains:
- rework class OvalNode
Fixes #37 
- rework functions evaluate of OVAL operators
- fixed negation evaluation in value node


matejak:
The PR seems to be coupled with #94. Next time, try to keep those two independent - now it makes sense to merge the former before reviewing this one.

Honny1:
@matejak The problem was solved.

oval_graph/oval_node.py
```diff
@@ -16,60 +16,69 @@ class OvalNode():
...
-        self.node_id = node_id
-        self.negation = self.validate_negation(input_negation)
-        self.node_type = self.validate_type(input_node_type)
-        self.value = self.validate_type_and_value(input_value)
+    def __init__(self, **kwargs):
+        try:
+            assert 'node_id' in kwargs
+            assert 'node_type' in kwargs
+            assert 'value' in kwargs
+            self.node_id = kwargs.get('node_id')
```
matejak: If you know that `node_id` is in `kwargs`, you can write `kwargs[node_id]`.

oval_graph/oval_node.py
```diff
@@ -16,60 +16,69 @@ class OvalNode():
...
+            self.node_type = self._validate_type(kwargs.get('node_type'))
+            self.value = self._validate_value(kwargs.get('value'))
+
+            self.negation = self._validate_negation(
+                kwargs.get('negation', False))
+        except AssertionError:
```
matejak: asserts may be dropped from the code if it is compiled with the `-O` flag. So you can write more general `except Exception as exc`, and then include the `str(exc)` message as an indication of what went wrong.

oval_graph/oval_node.py
```diff
@@ -16,60 +16,69 @@ class OvalNode():
...
+    def _validate_negation(self, input_negation):
+        assert isinstance(input_negation, bool)
         return input_negation
 
-    def validate_type(self, input_node_type):
+    def _validate_type(self, input_node_type):
         node_type = input_node_type.lower()
-        if node_type != "value" and node_type != "operator":
-            raise ValueError("err- unknown type")
+        assert node_type == "value" or node_type == "operator"
```
matejak: You can use `node_type in ("value", ...)`, and in this case, it is better to raise a [TypeError](https://docs.python.org/3/library/exceptions.html#TypeError).

oval_graph/oval_node.py
```diff
@@ -16,60 +16,69 @@ class OvalNode():
     '''
     The OvalNode object is one node of oval_graph.
 
-    Args:
+    Require args:
```
matejak: It's require**d**. You **require** them, therefore they are **required** by you.
And a small inspiration regarding docstrings - you can consider adopting the [Google format](https://www.sphinx-doc.org/en/master/usage/extensions/example_google.html#example-google), which is somehow standardized.

matejak:
LGTM, thanks for those improvements!

PR #97, opened 2020-03-05
Rework xml parser - split xml parser into smaller parts

Honny1:
This PR divides the `XmlParser` class by task into` _XmlParserComments`, `_XmlParserTestInfo`,` _XmlParserScanDefinitions`, `_BuilderOvalGraph`.
- `XmlParser` will process the xml file.
  1 - `_XmlParserTestInfo` creates information about each test from the processed file. Test information is assigned to the rule definition when the rule definition is created by ID._
  2 - `_XmlParserScanDefinitions` creates a dictionary of all rule definitions.
  3 - `_XmlParserComments` creates a dictionary with comments from the processed file and inserts comments into the definition dictionary. _Comment comment has the same layout as the definition dictionary._
  4 - `_BuilderOvalGraph` builds a rule graph from the rule definition dictionary.

### DIFF
--- a/oval_graph/_builder_oval_graph.py
+++ b/oval_graph/_builder_oval_graph.py
@@ -1,0 +1,53 @@
+import uuid
+
+from .oval_node import OvalNode
+
+# rework this (use static metodes)
+
+
+class _BuilderOvalGraph:
+    def __init__(self):
+        pass
+
+    def _definition_dict_to_node(self, dict_of_definition):
+        children = []
+        for child in dict_of_definition['node']:
+            if 'operator' in child and 'id':
+                children.append(self._definition_dict_to_node(child))
+            else:
+                children.append(
+                    OvalNode(
+                        node_id=child['value_id'],
+                        node_type='value',
+                        value=child['value'],
+                        negation=child['negate'],
+                        comment=child['comment'],
+                        tag=child['tag'],
+                        test_result_details=child['test_result_details'],
+                    ))
+
+        if 'id' in dict_of_definition:
+            children[0].node_id = dict_of_definition['id']
+            return children[0]
+        else:
+            return OvalNode(
+                node_id=str(uuid.uuid4()),
+                node_type='operator',
+                value=dict_of_definition['operator'],
+                negation=dict_of_definition['negate'],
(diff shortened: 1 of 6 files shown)


oval_graph/_xml_parser_oval_scan_definitions.py
```diff
@@ -30,7 +30,9 @@ def _build_graph(self, tree_data):
             graph['node'].append(self._build_node(tree, "Definition"))
         return graph
 
-    def _str_to_bool(self, s):
+    # rework this
```
matejak: Without further explanation, comments like this will confuse even the person who wrote them.

oval_graph/_builder_oval_graph.py
```diff
@@ -0,0 +1,53 @@
...
+                node_id=str(uuid.uuid4()),
+                node_type='operator',
+                value=dict_of_definition['operator'],
+                negation=dict_of_definition['negate'],
+                comment=dict_of_definition['comment'],
+                tag=dict_of_definition['tag'],
+                children=children,
+            )
+
+    def get_oval_graph_from_definition_dict_of_rule(self, rule):
```
matejak: The name is not so descriptive - you pass it the rule data structure, not the definition dict. Either rename the method, or make it accept the definition dict and the rule ID.

oval_graph/_xml_parser_test_info.py
```diff
@@ -0,0 +1,160 @@
...
+    def _find_item_ref(self, object_):
+        return list(
+            filter(
+                None, [
+                    self._get_item_ref(item) for item in object_]))
+
+    def _get_item_ref(self, item):
+        return item.get('item_ref') if item.get('item_ref') else None
+
+    def _get_unicate_key(self, key):
```
matejak: You probably want to call it `_get_unique_key` (not unicate, but unique).

oval_graph/_xml_parser_oval_scan_definitions.py
Honny1: This is my mark for rework in the next PR.

matejak:
Great fix, you have picked a good name. Merging, as those packit processes seem to be stuck.

PR #101, opened 2020-03-10
Rework xml parser scan definitions

Honny1:
This PR changes:
- internal arrays to dictionaries for sreaching performance improvement
- loading negation for nodes

### DIFF
--- a/oval_graph/_xml_parser_comments.py
+++ b/oval_graph/_xml_parser_comments.py
@@ -37,8 +37,7 @@ class _XmlParserComments:
     def _prepare_definition_comments(self):
         definitions = {}
         for definition in self.oval_definitions:
-            comment_definition = dict(
-                id=definition.get('id'), comment=None, node=[])
+            comment_definition = dict(comment=None, node=[])
             title = definition.find(
                 './/oval-definitions:metadata/oval-definitions:title', ns)
             description = definition.find(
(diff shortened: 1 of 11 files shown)


oval_graph/xml_parser.py
```diff
@@ -109,20 +108,16 @@ def _get_oval_definitions(self):
...
-            if rule['id_rule'] == rule_id:
-                return rule['id_def']
-        raise ValueError('err- 404 rule not found!')
+        if rule_id in self.used_rules:
+            return dict(rule_id=rule_id,
+                        definition_id=self.used_rules[rule_id]['id_def'],
+                        definition=self.scan_definitions[self.used_rules[rule_id]['id_def']])
+        elif rule_id in self.notselected_rules:
+            raise ValueError(
+                'err- rule "{}" was not selected, so there are no results.'
```
matejak: Let's make this more user-friendly and representative (in another PR) - current error messages are quite geeky, and the client should not show the traceback, but it should only print the message.

oval_graph/_xml_parser_comments.py
```diff
@@ -47,26 +46,25 @@ def _prepare_definition_comments(self):
...
-            definitions.append(comment_definition)
+            definitions[definition.get('id')] = comment_definition
         return definitions
 
     def _recursive_help_fill_comments(self, comments, nodes):
-        out = nodes
-        out['comment'] = comments['comment']
-        for node, comment in zip(out['node'], comments['node']):
+        for node, comment in zip(nodes, comments):
             node['comment'] = comment['comment']
```
matejak: Try to think about naming here - we have `node`, `comment`, `comment["node"]`, `node["node"]`, and even `comment["comment"]`, which is quite confusing. In graphs, you have nodes and edges (or relations), at the same time you have the data structure with comments, IDs, references etc., and you need to distinguish between them. Could you e.g. try to describe what is the node of a comment?

oval_graph/xml_parser.py
Honny1: I create issue for this.

matejak:
Much better, thanks!

PR #102, opened 2020-03-11
Rework builder of oval graph

Honny1:
This PR reworks class ```_BuilderOvalGraph```.


jan-cerny:
@Honny1 Looks good to me, great improvement! I also like very much the packit integration, I used the RPM provided by packit to check the changes. Thanks for setting this up!

PR #103, opened 2020-03-11
Rework xml parser for comments

Honny1:
This PR reworks class ```_XmlParserComments```.

### DIFF
--- a/oval_graph/_xml_parser_comments.py
+++ b/oval_graph/_xml_parser_comments.py
@@ -2,37 +2,43 @@ ns = {
     'oval-definitions': 'http://oval.mitre.org/XMLSchema/oval-definitions-5',
 }
 
-# rework this (use static metodes)
-
 
 class _XmlParserComments:
     def __init__(self, oval_definitions):
         self.oval_definitions = oval_definitions
 
-    def _create_dict_form_criteria(self, criteria, description):
+    def _create_dict_form_criteria(self, criteria, description=None):
         comments = dict(
-            operator='AND' if criteria.get('operator') is None else criteria.get('operator'),
-            comment=description if criteria.get('comment') is None else criteria.get('comment'),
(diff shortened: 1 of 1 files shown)


oval_graph/_xml_parser_comments.py
```diff
@@ -2,38 +2,44 @@
...
     def __init__(self, oval_definitions):
         self.oval_definitions = oval_definitions
 
-    def _create_dict_form_criteria(self, criteria, description):
+    def _create_dict_form_criteria(self, criteria, description=None):
         comments = dict(
-            operator='AND' if criteria.get('operator') is None else criteria.get('operator'),
-            comment=description if criteria.get('comment') is None else criteria.get('comment'),
+            operator=self._get_operator(criteria),
+            comment=self._get_comment(criteria, description),
```
jan-cerny, line 13: What happens if the description is None?

oval_graph/_xml_parser_comments.py
```diff
@@ -2,38 +2,44 @@
...
+        )
+        if criterion.get('definition_ref'):
+            out['extend_definition'] = criterion.get('definition_ref')
+        else:
+            out['value_id'] = criterion.get('test_ref')
+        return out
+
+    def _get_operator(self, criterion):
+        return 'AND' if criterion.get(
+            'operator') is None else criterion.get('operator')
```
jan-cerny, line 37: I'm not a big fan of this. I would prefer something like this:

```suggestion
        operator = criterion.get('operator')
        return 'AND' if operator is None else operator
```

oval_graph/_xml_parser_comments.py
Honny1: Returns ```None``` or ```criterion.get('comment')```.

jan-cerny:
Thanks for the updates, looks good to me.

PR #104, opened 2020-03-12
Rework client

Honny1:
This PR reworks the ```Client``` class because it contains methodes that can be simplified and more general. The files that are part of the html report have been split into smaller parts for future use in the html report with all the rules in one file.
- Simplified search rule metodes
- Simplified HTML creation metodes
- Split parts of HTML


jan-cerny:
The changes of code look fine. But I'm dissatisfied by the PR description. It says only that the patch reworks class Client. The description is missing information. Why is the class reworked? What are the benefits? Does it fix some problem? Are there any consequences? Answers to these questions should be part of any PR description.

Honny1:
This PR reworks the rule name search and file handling.

PR #108, opened 2020-04-01
Resolve variable references

Honny1:
This PR resolves references of variable  in OVAL details about test.

### DIFF
--- a/oval_graph/_xml_parser_test_info.py
+++ b/oval_graph/_xml_parser_test_info.py
@@ -83,7 +83,7 @@ class _XmlParserTestInfo:
         out = {}
         if len(collected_object) == 0:
             out[self._get_unique_id_in_dict(object_, out)
-                ] = self._get_object_items(object_)
+                ] = self._get_object_items(object_, collected_object)
         else:
             item_refs = self._find_item_ref(collected_object)
             if item_refs:
@@ -92,7 +92,7 @@ class _XmlParserTestInfo:
                         object_, out)] = self._get_item(item_id)
             else:
                 out[self._get_unique_id_in_dict(
-                    object_, out)] = self._get_object_items(object_)
+                    object_, out)] = self._get_object_items(object_, collected_object)
         return out
 
     def _xml_element_to_dict(self, object_, collected_object):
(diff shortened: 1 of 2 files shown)


jan-cerny:
Do you have some specific rule where we can see the advantages of this change? :smiley:

Honny1:
Yes, I have. Here are some examples.
```
xccdf_org.ssgproject.content_rule_audit_rules_unsuccessful_file_modification_chmod
xccdf_org.ssgproject.content_rule_audit_rules_unsuccessful_file_modification_openat_rule_order
```

jan-cerny:
Note: The variables handling is nicely visible with the second rule (xccdf_org.ssgproject.content_rule_audit_rules_unsuccessful_file_modification_openat_rule_order) after viewing test details of eg. oval:ssg-test_arufm_openat_order_32bit_eacces_augenrules:tst:1

oval_graph/_xml_parser_test_info.py
```diff
@@ -107,18 +107,37 @@ def _xml_element_to_dict(self, object_, collected_object):
...
+                if self._collected_object_is_not_none_and_not_empty_and_element_contain_var_ref(
+                        element, collected_object):
+                    variable_value = ''
+                    var_id = element.attrib.get('var_ref')
+                    for item in collected_object:
+                        if var_id == item.attrib.get('variable_id'):
+                            variable_value += item.text
+                        elif self._get_key_for_element(item) == 'message':
+                            variable_value = item.text[:item.text.find(
+                                '(') + 1] + var_id + ')'
```
jan-cerny, line 128: This feels clumsy. At least the index computation could be refactored out.

Also, what if there is a different message, with different meaning, perhaps completely unrelated?

oval_graph/_xml_parser_test_info.py
```diff
@@ -107,18 +107,37 @@ def _xml_element_to_dict(self, object_, collected_object):
...
+                else:
+                    out[self._get_unique_id_in_dict(element, out)] = 'no value'
         return out
 
+    def _collected_object_is_not_none_and_not_empty_and_element_contain_var_ref(
+            self, element, collected_object):
+        if collected_object is not None and 'var_ref' in element.attrib:
+            return len(collected_object)
+        return False
+
```
jan-cerny, line 140: The name of this function feels too long. Also, I think it's mixing 2 potentionally unrelated things. Personally I would not create a function for this. Or at least I would create a better name.

oval_graph/_xml_parser_test_info.py
```diff
@@ -107,18 +107,37 @@ def _xml_element_to_dict(self, object_, collected_object):
...
 
-    def _get_object_items(self, object_):
+    def _get_object_items(self, object_, collected_object):
         out = {}
         for element in object_.iterchildren():
             if element.text and element.text.strip():
                 out[self._get_unique_id_in_dict(element, out)] = element.text
             else:
-                out[self._get_unique_id_in_dict(element, out)] = "no value"
+                if self._collected_object_is_not_none_and_not_empty_and_element_contain_var_ref(
```
jan-cerny, line 119: Maybe you could move the whole else branch to a separate function. That can make the code more readable.

oval_graph/_xml_parser_test_info.py
Honny1: Messages are sometimes destroyed. The message is limited to 99 characters.

PR #109, opened 2020-04-01
Improve error messages

### LINKED ISSUE #106: Improve error messages
More user-friendly and representative - current error messages are quite geeky, and the client should not show the traceback, but it should only print the message.

Honny1:
This PR improves error message and error traceback for user.
Fixes #106

### DIFF
--- a/oval_graph/json_to_html.py
+++ b/oval_graph/json_to_html.py
@@ -43,13 +43,15 @@ class JsonToHtml(Client):
             try:
                 return json.load(f)
             except Exception as error:
-                raise ValueError("err- Used file is not json or valid.")
+                raise ValueError(
+                    'Used file "{}" is not valid json.'.format(
+                        self.source_filename))
 
     def load_json_to_oval_tree(self, rule):
         try:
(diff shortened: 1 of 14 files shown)


oval_graph/json_to_html.py
```diff
@@ -80,21 +82,16 @@ def _get_wanted_not_selected_rules(self):
...
+        out = []
+        for rule in rules["rules"]:
+            self.oval_tree = self.load_json_to_oval_tree(rule)
+            oval_tree_dict = self.create_dict_of_oval_node(self.oval_tree)
+            src = self.get_save_src(rule.replace('graph-of-', '') + "-")
+            self.save_html_report(oval_tree_dict, src)
+            self.open_web_browser(src)
+            print('Rule "{}" done!'.format(rule))
+            out.append(src)
+        return out
```
jan-cerny, line 94: This code looks similar to the code in `prepare_data` in `arf_to_html`. Would it be feasible to unify it?

Honny1: OK, I will remove this duplicate code in another PR.

PR #111, opened 2020-04-09
Fix processing negation

Honny1:
This PR fixes a problem with parsing negation. The problem is caused by saving the results of the negated test in ARF. Result of  negated test is negated whitch mean result of test has processed negation. When evaluating a tree, negation is processed. This results in double negation.

### DIFF
--- a/oval_graph/_xml_parser_oval_scan_definitions.py
+++ b/oval_graph/_xml_parser_oval_scan_definitions.py
@@ -28,12 +28,28 @@ class _XmlParserScanDefinitions:
             negate_status = str_to_bool[node.get('negate')]
         return negate_status
 
+    def _get_result(self, negate_status, tree):
+        """
+            This  method  removes  the  negation of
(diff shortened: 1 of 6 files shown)


oval_graph/_xml_parser_oval_scan_definitions.py
```diff
@@ -28,12 +28,23 @@ def _get_negate_status(self, node):
...
 
+    def _get_result(self, negate_status, tree):
+        result = tree.get('result')
+        reverse_negate_value = {
+            'true': 'false',
+            'false': 'true',
+        }
+        if negate_status and result in ('true', 'false'):
+            result = reverse_negate_value[result]
+        return result
```
jan-cerny, line 39: I believe you can improve this function by adding a comment. This function does something "unexpected" for a person that sees it for the first time. If the person sees that, they might think "WTH". A comment will help them to understand this quickly. Comment can be similar to the PR description of this PR.

PR #117, opened 2020-04-25
Integration tests

Honny1:
This PR creates command tests. These tests test the use of commands and the compatibility between the `arf-to-json` and `json-to-graph` commands.

### DIFF
--- a/tests/any_test_help.py
+++ b/tests/any_test_help.py
@@ -3,6 +3,8 @@ import json
 import sys
 import mock
 import pytest
+import uuid
+import tempfile
 
 from oval_graph.oval_node import restore_dict_to_tree, OvalNode
 from oval_graph.converter import Converter
@@ -36,6 +38,10 @@ def any_test_parsing_and_evaluate_scan_rule(src, rule_id, result):
     parser = XmlParser(get_src(src))
     oval_tree = parser.get_oval_tree(rule_id)
     any_test_treeEvaluation(oval_tree, result)
+
+
+def get_random_dir_in_tmp():
+    return os.path.join(tempfile.gettempdir(), str(uuid.uuid4()))
 
 
 def any_get_test_data_json(src):
--- a/tests/any_test_help.py
+++ b/tests/any_test_help.py
@@ -5,6 +5,7 @@ import mock
 import pytest
 import uuid
 import tempfile
+import re
 
 from oval_graph.oval_node import restore_dict_to_tree, OvalNode
 from oval_graph.converter import Converter
@@ -164,9 +165,12 @@ def compare_results_json(result):
     result = any_get_test_data_json(result)
     referenc_result = any_get_test_data_json(
         'test_data/referenc_result_data_json.json')
-    print(json.dumps(result))
-    assert result[list(result.keys())[
-        0]] == referenc_result["xccdf_org.ssgproject.content_rule_package_abrt_removed"]
+    rule_name = "xccdf_org.ssgproject.content_rule_package_abrt_removed"
+    result_rule_name = [
+        x for x in result.keys() if re.search(
+            rule_name, x)]
+    assert (result[result_rule_name[0]]
+            == referenc_result[rule_name])
 
 
 def get_questions_not_selected(capsys, client):
--- a/tests/test_command_arf_to_graph.py
+++ b/tests/test_command_arf_to_graph.py
@@ -1,0 +1,94 @@
+import subprocess
+import os
+import pytest
+from readchar import key
+
+import tests.any_test_help
+
+
+@pytest.mark.usefixtures("remove_generated_reports_in_root")
+def test_command_arf_to_graph():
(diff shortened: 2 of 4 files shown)


tests/test_command_arf_to_graph.py
```diff
@@ -0,0 +1,63 @@
...
+import pytest
+from readchar import key
+
+import tests.any_test_help
+
+
+@pytest.mark.usefixtures("remove_generated_reports_in_root")
+def test_command_arf_to_graph():
+    out = os.popen(
+        ('python3 -m oval_graph.command_line arf-to-graph --off-web-browser ' +
```
matejak, line 12: Don't use `os.system`, `os.spawn` or `popen` with string arguments. Use the subprocess module: https://docs.python.org/3/library/subprocess.html
The main reason for this is safety.

tests/any_test_help.py
```diff
@@ -158,9 +165,12 @@ def compare_results_json(result):
...
         'test_data/referenc_result_data_json.json')
-    print(json.dumps(result))
-    assert result[list(result.keys())[
-        0]] == referenc_result["xccdf_org.ssgproject.content_rule_package_abrt_removed"]
+    rule_name = "xccdf_org.ssgproject.content_rule_package_abrt_removed"
+    result_rule_name = [
+        x for x in result.keys() if re.search(
+            rule_name, x)]
+    assert result[result_rule_name[
+        0]] == referenc_result[rule_name]
```
matejak, line 173: It is better to use parentheses and break the line just before `==`, as one understands that what's on the previous line should equal to what's on the next line. Breaking just before an array index is less readable.

tests/test_command_arf_to_graph.py
```diff
@@ -0,0 +1,69 @@
...
+
+@pytest.mark.usefixtures("remove_generated_reports_in_root")
+def test_command_arf_to_graph():
+    out = subprocess.check_output(['python3',
+                                   '-m',
+                                   'oval_graph.command_line',
+                                   'arf-to-graph',
+                                   '--off-web-browser',
+                                   'tests/test_data/ssg-fedora-ds-arf.xml',
+                                   'xccdf_org.ssgproject.content_rule_package_abrt_removed'])
```
matejak, line 18: This is correct, as you pass the command as a list of arguments.

tests/test_command_arf_to_graph.py
```diff
@@ -0,0 +1,69 @@
...
+        out.decode('utf-8').split('\n')[-2])
+
+
+def test_command_arf_to_graph_with_out_parameter():
+    src = tests.any_test_help.get_random_dir_in_tmp()
+    subprocess.call(
+        ('python3 -m oval_graph.command_line arf-to-graph --off-web-browser ' +
+         '--out ' + src +
+         ' tests/test_data/ssg-fedora-ds-arf.xml xccdf_org.ssgproject.' +
+         'content_rule_package_abrt_removed'), shell=True)
```
matejak, line 29: This is not correct, because you pass a single string, and then you have to specify `shell=True`, so the string is decomposed into the command and its arguments. You can fix this easily, as `subprocess.call` accepts a list/tuple of command components.

tests/test_command_arf_to_graph.py
```diff
@@ -0,0 +1,69 @@
...
+    tests.any_test_help.compare_results_html(
+        os.path.join(src, os.listdir(src)[0]))
+
+
+def test_inquirer_choice_rule():
+    src = tests.any_test_help.get_random_dir_in_tmp()
+    sut = pexpect.spawn(
```
matejak, line 36: `pexpect.spawn` also accepts a command and its args separately.

tests/test_command_arf_to_graph.py
```diff
@@ -0,0 +1,69 @@
+import subprocess
+import os
+import pexpect
```
matejak, line 3: If `pexpect` is not available, tests that require it should be skipped, as you have majority of tests that don't require `pexpect`. Check out `pytest.skipif`: https://docs.pytest.org/en/latest/skipping.html#id1

matejak:
Sorry, still not satisfied, but you are getting close.
I have also noticed that you have introduced the `pexpect` dependency, so don't forget to document it, and as it is a non-essential dependency, allow partial execution of your test suite if it is missing.

tests/test_command_arf_to_json.py
```diff
@@ -0,0 +1,127 @@
...
+            'oval_graph.command_line',
+            'arf-to-json',
+            'tests/test_data/ssg-fedora-ds-arf.xml',
+            'xccdf_org.ssgproject.content_rule_package_abrt_removed'
+        ])
+    tests.any_test_help.compare_results_json(src)
+
+
+def test_inquirer_choice_rule():
+    pexpect = pytest.importorskip("pexpect")
```
matejak, line 46: Very nice, I like this!

matejak:
Good job, merging!

PR #124, opened 2020-07-27
Fixes #116

### LINKED ISSUE #116: Processing only filed rules
Automatic processing if one rule is found. Not work when is used parameter `--show-failed-rules`. And if the `--all` parameter is used with a combination of the ` --show-failed-rules` parameter, the command will process all the rules. I think a command with these parameters should handle all failed rules.
```
python3 -m oval_graph.command_line arf-to-graph --show-failed-rules  tests/test_data/ssg-fedora-ds-arf.xml "_package_\w+_removed"
```

Honny1:
This PR fixes problem with combination of parameters ```--all``` and ```--show-filed-rules```. 
More info about problem is in #116. 
Fixes: #116

### DIFF
--- a/oval_graph/client.py
+++ b/oval_graph/client.py
@@ -44,7 +44,7 @@ class Client():
     def run_gui_and_return_answers(self):
         if self.isatty:
             if self.all_rules:
-                return {'rules': self.search_rules_id()}
+                return self._get_rules()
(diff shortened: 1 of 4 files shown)


matejak:
LGTM, thank you for the PR!

PR #129, opened 2020-08-20
Code documentation

Honny1:
This PR contains:
- Documentation strings about the contents of the file where I think is helpful.
- Disables warning missing docstring about contents of the file in pylint config.  
- Renames Class BuilderOvalGraph to BuilderOvalTree.

### DIFF
--- a/oval_graph/__init__.py
+++ b/oval_graph/__init__.py
@@ -1,1 +1,5 @@
+"""
+    Tool for visualizing OVAL rules from
+    ARF XML generated with SCAP scanners.
+"""
 __version__ = '1.2.5'
--- a/oval_graph/command_line.py
+++ b/oval_graph/command_line.py
@@ -1,3 +1,7 @@
+"""
+    This file contains entry points for commands
+"""
+
 import sys
 import traceback
 
--- a/oval_graph/converter.py
+++ b/oval_graph/converter.py
@@ -1,3 +1,7 @@
+"""
+    This file contains a class for converting OVAL tree to JSON used in HTML
+"""
+
 import re
 
 from .oval_node import OvalNode
--- a/oval_graph/oval_node.py
+++ b/oval_graph/oval_node.py
@@ -1,13 +1,8 @@
 import oval_graph.evaluate
 
-'''
-    This module contains methods and classes for
-    constructing and controlling an oval tree.
-'''
-
 
 class OvalNode():
-    '''
+    """
     The OvalNode object is one node of oval_graph.
 
     Attributes:
@@ -21,7 +16,7 @@ class OvalNode():
         OVAL definition or XCCDF rule
         test_result_details (dict|None): information about test
         children ([OvalNode]): children of node
-    '''
+    """
 
     def __init__(self, **kwargs):
         """
--- a/oval_graph/xml_parser.py
+++ b/oval_graph/xml_parser.py
@@ -7,7 +7,7 @@ import sys
 
 from lxml import etree as ET
 
-from ._builder_oval_graph import _BuilderOvalGraph
+from ._builder_oval_tree import _BuilderOvalTree
 from ._xml_parser_oval_scan_definitions import _XmlParserScanDefinitions
 from .exceptions import NotChecked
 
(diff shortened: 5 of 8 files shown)


oval_graph/_builder_oval_graph.py
```diff
@@ -1,3 +1,7 @@
+'''
+    This file contains class for building oval graph
```
jan-cerny, line 2:
```suggestion
    This file contains class for building OVAL graph
```

oval_graph/converter.py
```diff
@@ -1,3 +1,7 @@
+'''
+    This file contains a class for convertinging oval tree to JSON used in HTML
```
jan-cerny, line 2:
```suggestion
    This file contains a class for converting OVAL tree to JSON used in HTML
```

oval_graph/command_line.py
```diff
@@ -1,3 +1,7 @@
+'''
+    This file contains a entry points for commands
```
jan-cerny, line 2:
```suggestion
    This file contains entry points for commands
```

oval_graph/exceptions.py
```diff
@@ -1,3 +1,7 @@
+'''
+    This file contains a custom expection
```
jan-cerny, line 2:
```suggestion
    This file contains a custom exception
```

oval_graph/oval_node.py
```diff
@@ -1,17 +1,11 @@
 '''
-    Modules form my lib and for create ID
+    This file contains a oval node class
```
jan-cerny, line 2: Do you really think that this will be helpful for anyone? We can see that there is OvalNode class in this file. You can describe it more.

oval_graph/xml_parser.py
```diff
@@ -1,3 +1,7 @@
+'''
+    This file contains a class for processing oval graph from XML
```
jan-cerny, line 2:
```suggestion
    This file contains a class for creating OVAL graph from XML source
```

oval_graph/__init__.py
```diff
@@ -1 +1,8 @@
+'''
+    This module contains methods and classes for
+    constructing and controlling an oval tree and
```
jan-cerny, line 3:
```suggestion
    constructing and controlling an OVAL tree and
```

jan-cerny (commented):
The PR title suggests that the PR adds documentation strings but actually it moves a lot of code around and performs other changes than in the documentation.

oval_graph/oval_node.py
matejak: Or, conversely, one doesn't have to describe it at all. The problem with documentation is that it may not get updated as frequently as the code, so it may become misleading.

matejak:
I know that various checkers report errors if something is not documented, but they don't report errors if the documentation is useless (because it is trivial), if it is outdated, or even if it is just wrong.
I would go for the "less is more" approach - if a method has a good name, it quite often doesn't need a docstring, and it is similar to modules that contain class definitions - you probably don't need to document what is the purpose of the `exceptions` module.

oval_graph/command_line_client/__init__.py
```diff
@@ -0,0 +1,3 @@
+'''
+    This module is intended for processing command parameters.
```
matejak, line 2: You probably don't mean *module*, but a *package*. But again, I wouldn't document this, one can find out what is where quickly with low effort, and without risk of using outdated information.

oval_graph/_builder_html_graph.py
```diff
@@ -1,3 +1,7 @@
+'''
+    This file contains class for building HTML with graph
```
matejak, line 2: The file contains a class called `BuilderHtmlGraph` - the name already indicates that it builds a HTML with the graph.

matejak:
There is nothing clearly wrong with documentation, but as [Rober C Martin would say](https://medium.com/@huytrongnguyen1985/lessons-learnt-from-the-clean-code-robert-c-martin-cecbe2b09139), one should document public interfaces and quirks in the code that are caused by limitations of the programming language or of the chosen framework.
Almost all other code documentation can be avoided correctly naming functions and variables, and by keeping those functions/methods small. I think that majority of the documentation that you propose doesn't fix any real problem - good for you! As you use descriptive names, it just states the obvious.

matejak:
I like the evolution of the PR, it led to massive simplification of test suite output, and the code is readable even without extensive comments due to good naming practices.

PR #130, opened 2020-08-20
Remove unnecessary code to create html report

### LINKED ISSUE #122: Use same code to create report with one rule.
* Remove unnecessary code to create html report. Use the code that is used when ```--all-in-one``` is applied.
* Append a title with the rule name. As in the ```--all-in-one``` case.

Honny1:
This PR removes unnecessary code to create HTML. And it uses the same code as for the `--all-in-one` option. A title with the name of the rule was added above the graph, because the same code is used as in the case of `--all-in-one`.

Fixes #122


oval_graph/client.py
```diff
@@ -242,21 +240,25 @@ def _merge_report_parts(self, data):
...
+            "\n<script>var data_of_tree = " + str(
+                json.dumps(
+                    {
+                        re.sub(
+                            r'\-\d{2}\_\d{2}\_\d{4}\-\d{2}\_\d{2}\_\d{2}$',
+                            '',
+                            k): v for k,
+                        v in dict_.items()},
+                    sort_keys=False,
+                    indent=4)) + ";</script>\n")
```
matejak, line 254: It is totally unclear what this code is doing. `data` and `dict_` are generic names, and the regular expression is complicated. Consider extracting this functionality into a named private method, and naming the regexp by storing it in a variable.

Aside from that, you are creating contents of an HTML document, so you could even consider using the lxml functionality, and create parts of it programatically.

oval_graph/json_to_html.py
```diff
@@ -92,20 +92,11 @@ def create_dict_of_rule(self, rule):
         return self.create_dict_of_oval_node(self.oval_tree)
 
     def _put_to_dict_oval_trees(self, dict_oval_trees, rule, date=None):
-        dict_oval_trees[rule] = self.create_dict_of_rule(rule)
+        dict_oval_trees[rule.replace(
+            'graph-of-', '')] = self.create_dict_of_rule(rule)
```
matejak, line 96: `graph-of-` looks like a constant, so please declare it somewhere, and don't repeat the string.

matejak (changes requested):
This PR introduces numerous improvements that simplify the code, and there is a nice opportunity to input even more improvements towards greater maintainability.

oval_graph/client.py
```diff
@@ -182,83 +182,116 @@ def get_src(self, src):
...
+        )
+
+    def _get_html_body(self, dict_of_rules):
+        return E.body(
+            E.script(self._get_script_graph_data(dict_of_rules)),
+            self._get_titles_and_places_for_graph(dict_of_rules),
+            E.div({'id': 'data'}),
+            E.div({'id': 'modal', 'class': 'modal'},
+                  E.div({'class': 'modal-content'},
+                        E.span({'id': 'closer', 'class': 'close'}, '&times;'),
```
matejak, line 244: I don't understand why, but I guess that this is not correct - when I click rule details, there is not the "X", but literal `&times;` that doesn't close the window if clicked.

matejak (changes requested):
A regression got there: When I click test details, there is not the "X" to close the window, but there is literal `&times;` that can't be clicked.

oval_graph/client.py
```diff
@@ -182,83 +182,116 @@ def get_src(self, src):
...
-                oval_tree_dict = self.create_dict_of_rule(rule)
-                src = self._get_src_for_one_graph(rule, date)
-                self.save_html_and_open_html(
-                    oval_tree_dict, src, rule, out)
-            except NotChecked as error:
-                self.print_red_text(error)
+    def prepare_data(self, rules):
+        out = []
+        oval_tree_dict = dict()
+        date = str(datetime.now().strftime("-%d_%m_%Y-%H_%M_%S"))
```
matejak, line 205: and while we are at it, this format of time is used project-wide, so I would also suggest to make it a constant. Not necessarily in this PR though.

Honny1: I'll fix it in #131

oval_graph/client.py
```diff
@@ -182,83 +182,116 @@ def get_src(self, src):
...
-                         rule.split('-')[2] +
-                         '</h1>' +
-                         '<div id="' +
-                         re.sub(r'[\_\-\.]', "", rule) +
-                         '"></div>')
+            for line in data_file.readlines():
+                out += line
+            return out
+
+    def _get_html_head(self):
```
matejak, line 226: That's much better!

matejak (approved):

matejak:
Works great, thank you for the PR!

PR #134, opened 2020-09-02
Silent mode

### LINKED ISSUE #132: Utilities pring verbose-class output by default.
Tools should obey the [silent by default](http://www.linfo.org/rule_of_silence.html) convention, e.g. so it is more convenient to wrap them.

Right now, I see messages `Results are saved: ...` - those are fine for verbose/debug output, but they shouldn't be displayed by default.

Honny1:
This PR applies the rule of silence.
* renames --off-web-browser to --display
* reworks enabling and disabling the web browser
* changes default output directory to ```/tmp```

Fixes #132

### DIFF
--- a/oval_graph/command_line.py
+++ b/oval_graph/command_line.py
@@ -10,9 +10,9 @@ CEND = '\033[0m'
 
 def print_where_is_saved_result(results_src):
     if results_src:
-        print("Results are saved:")
+        print("Results are saved:", file=sys.stderr)
         for src in results_src:
-            print(src)
+            print(src, file=sys.stderr)
 
 
 def arf_to_graph(args=None):
@@ -54,7 +54,8 @@ def main(client):
             results_src = client.prepare_data(answers)
     else:
         results_src = client.prepare_data({'rules': [rules[0]]})
-    print_where_is_saved_result(results_src)
+    if client.verbose:
+        print_where_is_saved_result(results_src)
 
 
 if __name__ == '__main__':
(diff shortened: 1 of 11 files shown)


oval_graph/command_line.py
```diff
@@ -8,13 +8,6 @@
 CEND = '\033[0m'
 
 
-def print_where_is_saved_result(results_src):
```
matejak, line 11: So don't throw away this verbose code, just make it optional.

oval_graph/command_line.py
```diff
@@ -8,13 +8,6 @@
...
-    if results_src:
-        print("Results are saved:")
```
matejak, line 13: If a program outputs data to be processed, it usually does that to standard output. If it displays its internal log, it usually uses standard error. So I would suggest using `print(..., file=sys.stderr)` instead.

matejak (changes requested):
The PR works great, I just have some remarks:

- What happens when one runs the script on an ARF without specifying `-o` or `--on-web-browser`? Now it looks like that nothing will happen - no file is saved, and nothing is displayed. I think that it is reasonable to display the browser if no output file is selected.
- The `--on-web-browser` is a bit clumsy. I would advise to use something like `--display`, or `--show`.
- It is OK to display details in the console if the user asks for it - you can have verbose output, or even debug output if you want. This may help you or your users when they hit a bug and you try to find out what is going on. So don't throw away the old verbose code, just make it optional.

matejak:
LGTM, great work!

matejak (approved):

PR #136, opened 2020-09-04
Build script

Honny1:
This PR includes:
* fixes a script build issue
* runs test before build
* prepares the build environment
* rework build script
* change PackIt relase config (Because when you create a release, PackIt creates too many unused PR into the RPM upstream.)

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,41 +1,89 @@
-set -e 
+#!/bin/sh
+# Build script of python package
+
+set -e
 name="oval-graph"
 module="oval_graph"
 
-if [ "$1" != "" ]; then
-    new_version=$1
-    if [[ $new_version =~ ^[0-9]{1,}.[0-9]{1,}.[0-9]{1,}$ ]]; then
-        old_version=$(python3 setup.py --version)
-        git checkout master
-        git pull upstream master
-        if git status --porcelain=v1 | grep -q '^\(.M\|M.\)'; then
-            echo "Not commited changes!"
-        else
-            # update version in file
-            sed -i "s/$old_version/$new_version/g" ${module}/__init__.py
-            version=$(python3 setup.py --version)
-            rpmdev-bumpspec oval-graph.spec --comment="release ${version}" -n "${version}" --userstring="Jan Rodak <jrodak@redhat.com>"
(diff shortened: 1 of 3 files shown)


build.sh
```diff
@@ -1,37 +1,77 @@
...
-                git add oval-graph.spec   
-                git commit -m "${version}"
-                git tag "${version}"
-                git push --follow-tags
-                git push upstream master
-                # Build  
-                rm -rf dist
-                python3 setup.py pytest sdist bdist_wheel
-                twine upload dist/*
+                commit_version "$version" "$module"
```
matejak, line 63: That would be the last thing to do - first make the change to the version, then go all the way to tests, then commit, and only when you feel confident push the tag. You can undo pushing a tag, but that should be an absolute exception.

build.sh
```diff
@@ -1,37 +1,77 @@
...
+
+commit_version() {
+    version=$1
+    module=$2
+    git add ${module}/__init__.py
+    git add oval-graph.spec
+    git commit -m "${version}"
+    git tag "${version}"
+    git push --follow-tags
+    git push upstream master
```
matejak, line 20: My suggestion (i.e. not merge condition) would be to perform those things manually, maybe have the script print those commands, so it is easy to execute them.
~You want to be sure that you didn't forget to commit a file, or that unwanted files are not committed, run `git status` before committing etc.~ I see that you already do that, that's cool.

build.sh
```diff
@@ -1,37 +1,77 @@
...
+}
+
+load_upstream() {
+    git checkout master
+    git pull upstream master
+}
+
 if [ "$1" != "" ]; then
     new_version=$1
-    if [[ $new_version =~ ^[0-9]{1,}.[0-9]{1,}.[0-9]{1,}$ ]]; then
```
matejak, line 7: Those nested `if`s are difficult to read. You can use `exit` or define and use a `die` function to simply stop the script if something is not OK. That way, you won't have to have an `else` block.

matejak (changes requested):
I like general additions to the script and the improved readability, and I think that there is still room for improvements.

matejak:
Great improvements, merging!

matejak (approved):

PR #137, opened 2020-09-07
Creates a shorter option for command arguments

Honny1:
This PR creates a shorter option for command arguments.

### DIFF
--- a/oval_graph/client.py
+++ b/oval_graph/client.py
@@ -212,7 +212,7 @@ class Client():
             default=False,
             help=(
                 "Do not display passing tests for better orientation in"
-                " graphs that contain a large amount of nodes.(Not implemented)"))
+                " graphs that contain a large amount of nodes."))
         self.parser.add_argument(
             '-v',
             '--verbose',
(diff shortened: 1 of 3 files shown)


oval_graph/client.py
```diff
@@ -201,18 +201,21 @@ def prepare_parser(self):
...
             default=False,
             help="Process all matched rules.")
         self.parser.add_argument(
+            '-p',
             '--hide-passing-tests',
             action="store_true",
             default=False,
             help=(
                 "Do not display passing tests for better orientation in"
                 " graphs that contain a large amount of nodes.(Not implemented)"))
```
matejak, line 216: If it is not implemented, don't have it here, just comment it out.

matejak (changes requested):
Not all arguments need a long and short option - I recommend to implement this for ones that you expect to be used most. But the proposal is OK.

I have noticed that you have a duplication in arguments parsing code - I recommend to share the code, so consistency of interfaces is maintained automatically.

oval_graph/client.py
Honny1: This option has been implemented, but I forgot to remove it.

matejak:
LGTM, merging, I will create issues for the duplication.

matejak (approved):

PR #140, opened 2020-09-09
Improves performance when creating HTML

Honny1:
This PR improves the HTML creation process. Reduces reads from the hard disk and speeds up the saving of generated HTMLs.

Edit: You can test improvement with benchmark script named  `benchmark_full_command.py`.

### DIFF
--- a/oval_graph/client.py
+++ b/oval_graph/client.py
@@ -35,6 +35,7 @@ class Client():
         self.START_OF_FILE_NAME = 'graph-of-'
         self.date = str(datetime.now().strftime("-%d_%m_%Y-%H_%M_%S"))
         self.verbose = self.arg.verbose
+        self.html_builder = None
 
     def _get_message(self):
         MESSAGES = {
@@ -166,9 +167,7 @@ class Client():
         return str(os.path.join(_dir, src))
 
     def _build_and_save_html(self, dict_oval_trees, src, rules, out_src):
-        builder = BuilderHtmlGraph(
-            self.parts, self.verbose)
-        builder.save_html(dict_oval_trees, src, rules)
+        self.html_builder.save_html(dict_oval_trees, src, rules)
         out_src.append(src)
 
     def open_html(self, out):
--- a/oval_graph/client.py
+++ b/oval_graph/client.py
@@ -3,14 +3,11 @@ import argparse
 import tempfile
 import os
 import webbrowser
-import json
-import shutil
 from datetime import datetime
 import sys
 
 from .xml_parser import XmlParser
 from .exceptions import NotChecked
-from ._builder_html_graph import BuilderHtmlGraph
 from .__init__ import __version__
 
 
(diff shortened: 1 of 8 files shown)


matejak:
When I run `benchmark.py` before and after, the result is almost the same. How can one see the improvement? The new benchmark script doesn't do the same thing on master and this branch.

What is the nature of the speed improvement? When I checked out the source code, I haven't seen changes leading to disk access reduction. Could you be more specific?

Finally, I see that the `html_builder` property is defined even for classes that don't use it, e.g. `JsonToHtml`, `ArfToHtml`. What is the reason for that?

Honny1:
You are using a bad benchmark script. I should have written it in the description. Benchmark script which testing generating and saving HTML si named `benchmark_full_command.py`. Here are the results of `benchmark_full_command.py`: 
Before:
image
After:
image

I think merging these benchmark scrips will be a good idea.

The biggest performance change is caused by using a binary file and not wasting time converting an HTML object to text for a regular file.

The `html_builder` variable is defined in the `Client` class because the `Client` class is an intersection of the `ArfToHtml`, `ArfToJson`, and `JsonToHtml` classes to remove duplicate code. Defining an ```html_builder``` in ```__init__``` saves time when generating multiple HTML files, because the `html_builder` prepares repetitive parts of the html (head, scripts, etc..) already in  `__init__` of `HtmlBulder`.

oval_graph/client.py
```diff
@@ -35,6 +35,7 @@ def __init__(self, args):
         self.START_OF_FILE_NAME = 'graph-of-'
         self.date = str(datetime.now().strftime("-%d_%m_%Y-%H_%M_%S"))
         self.verbose = self.arg.verbose
+        self.html_builder = None
```
matejak, line 38: Why not to do `self.html_builder = BuilderHtmlGraph(self.parts, self.verbose)` here? We clearly need a `self.html_builder` in every `Client` instance.

matejak:
Thanks for the clarification - there is clearly an overhead with creation of the builder, and drop of the encoding and disk IO may help as well. I have just one suggestion.

oval_graph/client.py
```diff
@@ -23,18 +21,40 @@ def __init__(self, args):
...
+        if type(self).__name__ in ('ArfToHtml', 'ArfToJson'):
+            self._set_attributes_for_ArfToHtml_or_ArfToJson()
+
+        self.xml_parser = None
+        if type(self).__name__ != 'JsonToHtml':
+            self.xml_parser = XmlParser(self.source_filename)
+
+        self.html_builder = None
+        if type(self).__name__ != 'ArfToJson':
+            self.html_builder = BuilderHtmlGraph(self.parts, self.verbose)
```
matejak, line 48: This is an unfortunate construct - in object-oriented programming, the base class shouldn't care who derives from it.

oval_graph/client.py
```diff
@@ -23,18 +21,40 @@ def __init__(self, args):
...
+
+    def _set_attributes_for_ArfToHtml_or_JsonToHtml(self):
```
matejak, line 50: Again, it would be better to have a `_set_attributes` method in different derived classes.

oval_graph/client.py
```diff
@@ -23,18 +21,40 @@ def __init__(self, args):
...
+        self.all_in_one = self.arg.all_in_one
+        self.all_rules = True if self.all_in_one else self.arg.all
+        self.display_html = True if self.out is None else self.arg.display
+
+    def _set_attributes_for_ArfToHtml_or_ArfToJson(self):
```
matejak, line 55: Is the name correct? Shouldn't there be `ArfToJson`

matejak:
Works for me as an intermediate step towards a somehow cleaner object design. Good job!

PR #142, opened 2020-09-15
Embed search bar

### LINKED ISSUE #121: Embed Search bar
Embed Search bar to HTML report with graphs when is used option ```--all-in-one```.

Honny1:
This PR embeds search bar to HTML with graphs, when there are two or more graphs in one file.
(When is used parameter ``--all-in-one``)
Fixes #121


matejak:
Works great, good job!

PR #146, opened 2020-09-29
Fix duplicated definitions of command-line parsers

### LINKED ISSUE #145: Duplicated definitions of command-line parsers
The command-line utilities share the command-line interface, but the parsers are defined separately, i.e. they are not shared by utilities that use them. Check it out:

- https://github.com/OpenSCAP/OVAL-visualization-as-graph/blob/master/oval_graph/arf_to_html.py#L33
- https://github.com/OpenSCAP/OVAL-visualization-as-graph/blob/master/oval_graph/json_to_html.py#L88
- https://github.com/OpenSCAP/OVAL-visualization-as-graph/blob/master/oval_graph/arf_to_json.py#L61

This kind of duplication allows introduction of inconsistencies e.g. when fixing typos, or when you decide to change an interface for one utility and forget to make the same change for the other one.
I recommend use functions or mixin classes to remove the duplication.

Honny1:
This PR fixes duplicated definitions of command-line parsers.
Fixes #145

### DIFF
--- a/oval_graph/arf_to_html.py
+++ b/oval_graph/arf_to_html.py
@@ -32,25 +32,5 @@ class ArfToHtml(Client):
 
     def prepare_parser(self):
         super().prepare_parser()
-        self.parser.add_argument(
-            '-i',
-            '--all-in-one',
-            action="store_true",
-            default=False,
-            help="Processes all rules into one file.")
-        self.parser.add_argument(
-            '-d',
-            '--display',
-            action="store_true",
-            default=False,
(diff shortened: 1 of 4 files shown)


matejak:
Nice, there is just one small issue - the established way of naming is that when a function/method starts with `get`, then it should return something. You are quite close and obviously one can deviate from this convention, but you already named the parent method `prepare_parser`, so you can rename `get_args_when_output_is_html` to `prepare_args_when_output_is_html` and so on - that would me much more accurate.

matejak:
I like it, merging!

PR #147, opened 2020-10-01
Creates an --version argument

Honny1:
This PR creates an --version argument.


matejak:
Very good, the version message is the same for all utilities.
And has `oval-graph` became the official name of the tool?

PR #151, opened 2020-10-14
Moves the open web browser to the client

Honny1:
This PR moves function for opening the web browser to the client because opening HTML will be used for future command `arf-to-report`.


matejak:
LGTM, works, merging!

PR #152, opened 2020-10-20
Print traceback with verbose

Honny1:
This PR implements print traceback with the verbose parameter. For better-debugging experience.

### DIFF
--- a/oval_graph/command_line.py
+++ b/oval_graph/command_line.py
@@ -1,4 +1,5 @@
 import sys
+import traceback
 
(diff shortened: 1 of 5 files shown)


matejak:
The code is consistent with the existing code, but relying on `sys.argv` to have correct contents in the middle of the program is risky - any operation can change `sys.argv` contents. In other words, a solution without a global variable would be better.

Anyway, this is a separate issue, so I merge the PR, nice addition!

PR #156, opened 2020-11-06
Rework client

Honny1:
This PR reworks Client classes:

- Moves Client parts to a separate directory for better package organisation 
- Reworks the Client class for a purpose to atomise the Client class and remove code duplication
- Creates Client subclasses according to input and output of commands  
- Reworks classes for processing commands based on rework Client class 
- Moves Client tests to a separate directory for better tests organisation 
- Modifies tests based on rework


matejak:
Please try to explain what went on in greater detail than that classes were redesigned to reduce duplication.

From what I see, files were moved around, but the class structure was changed as well. Moreover, it all took place in one commit (Reworks and organizes the client), so it's difficult to identify logical steps that this PR consists of.

Honny1:
@matejak  I updated the description of PR and separated some changes to more commits.

oval_graph/command_line_client/client.py
```diff
@@ -1,93 +1,95 @@
...
-import subprocess
-from datetime import datetime
+import re
 import sys
+from datetime import datetime
 
-from .xml_parser import XmlParser
-from .exceptions import NotChecked
-from .__init__ import __version__
+from ..__init__ import __version__
```
matejak, line 6: ` from .. import __version__` should work as well - contents of `__init__.py` become part of the respective python *package*

matejak:
It is much better after you have split those changes to separate commits.

Next time, you can be more concrete - you have split the `Client` class into

- `ClientArfInput`,
- `ClientJsonInput` and
- `ClientHtmlOutput`

The HTML output can be viewed in a browser, so the class contains methods to handle browser interaction.

I have following questions:

- How come that I don't see `prepare_data` defined in `ClientJsonInput`, although it is an abstract method of `Client`?
- Why is `print_red_text` used only in the HTML output?

and one remark - the XML parser seems to be in fact an ARF XML parser, therefore I recommend to consider expanding this PR with the rename, to make it more obvious.

Honny1:
@matejak  
Metode `prepare_data` is defined in subclass `ClientHtmlOutput`. 
In class `ClientJsonImput` is not defined because it causes import problems whit overwriting metode in class `JsonToHtml`.
Classes `ArfToHtml`, `ArfToJson` and `JsonToHtml` are end classes for commands `arf-to-html`, `arf-to-json` and `json-to-html`.
Here is the Class diagram. 

Is this class design correct?
I also plan to rework the `XmlParser` class.

matejak:
>  Metode `prepare_data` is defined in subclass `ClientHtmlOutput`.

Thanks for the diagram, it helps to clarify the layout.
Generally, when you declare an abstract method, you expect that it is useful to all derived classes, it just may not be clear how exactly, and those derived classes should implement that method. In your case, e.g. `ClientJsonInput` doesn't implement that method. This indicates that not all clients find it useful, so I would suggest not to have it declared like this in the base `Client` class.

matejak:
This is a huge PR, and it improves the architecture of the project mainly by introducing inheritance to eliminate code duplication. Extensive test suite exists exactly for occasions like this - it provides great confidence that the project is functional even after such large-scale refactoring.
Well done, there are no more obvious things to improve, therefore I am merging this PR.

PR #157, opened 2020-11-18
Removes the use of sys.argv in the project

### LINKED ISSUE #154: Don't use sys.argv in the project
This variable shares all risks of using global variables - other parts of the program may change their contents.
A better approach than "this program has been called with `-v` or `--verbose`" is "the program knows, that the `verbose_mode` is `true`".

Solution of https://github.com/OpenSCAP/oval-graph/blob/master/oval_graph/command_line.py#L73 can be implemented using [argparse subcommands](https://docs.python.org/3/library/argparse.html#sub-commands) together with [parse_known_args](https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.parse_known_args).

Honny1:
This PR removes the use of  ```sys.argv``` in the project. 

Fixes #154

### DIFF
--- a/oval_graph/command_line.py
+++ b/oval_graph/command_line.py
@@ -17,41 +17,29 @@ def print_where_is_saved_result(results_src):
 
 
 def print_detail_traceback_if_verbose(args):
-    if args is None:
-        args = sys.argv[1:]
-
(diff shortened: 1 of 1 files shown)


matejak:
Perfect! Great work!

PR #163, opened 2020-12-17
Fix loading of ARF results when comment node is missing.

### LINKED ISSUE #162: Problem loading arf results
There is some problem in the following line that `node` key is not found for `node_of_comment`.

https://github.com/OpenSCAP/oval-graph/blob/0c0ea1024ee3bf45f405cf7f13018b9b235afbc3/oval_graph/_xml_parser_comments.py#L63

If we add a check in the `if` statement `and node_of_comment.get('node')`, it prevents from erroring and show the graph, but I don't know exactly what's the impact.

The code change I made:
```
                if 'operator' in node and node_of_comment.get('node'):
                    self._recursive_help_fill_comments(
                        node_of_comment['node'], node['node'])
```

ARF Results file that causes the issue: [xccdf_org.ssgproject.content_profile_anssi_bp28_high-xccdf-arf-results.xml.txt](https://github.com/OpenSCAP/oval-graph/files/5709976/xccdf_org.ssgproject.content_profile_anssi_bp28_high-xccdf-arf-results.xml.txt) (I had to rename the file with extension `.txt` so github let me upload this file)

ggbecker:
Fixes #162

### DIFF
--- a/oval_graph/_xml_parser_comments.py
+++ b/oval_graph/_xml_parser_comments.py
@@ -58,7 +58,7 @@ class _XmlParserComments:
     def _recursive_help_fill_comments(self, tree_of_comments, tree):
         for node, node_of_comment in zip(tree, tree_of_comments):
             node['comment'] = node_of_comment['comment']
-            if 'operator' in node:
+            if 'operator' in node and node_of_comment.get('node'):
                 self._recursive_help_fill_comments(
                     node_of_comment['node'], node['node'])
 


ggbecker:
I suspect that this issue happens because of missing comments in the criteria section, which in this case was: https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/bootloader-grub2/non-uefi/grub2_password/oval/shared.xml#L6 and https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/bootloader-grub2/non-uefi/grub2_password/oval/shared.xml#L8

Honny1:
Works great, merging. Thank you for the fix!

ggbecker:
> One thing that happens with this fix is that a comment from a different part of the XML gets into the criteria that misses the comment.

@Honny1 I still think that this fix was not the best solution for the problem, the comments attribute is being shuffled among nodes if comments are missing. A proper solution might be needed to handle missing comments.

Honny1:
@ggbecker  I'll take a closer look at a problem. I think where could be a problem.

Honny1:
@ggbecker I took a closer look at a problem. I visualise rule that you mentioned and this is a result.
image
Oval-graph displayed the title of the definition at the first node and a description of the definition is displayed on the second node (top criteria node). The second node is usually without comment so I decide to use empty place to display more information. Here is part of arf were I reading comments, title and description of definition rule.
```
<oval-def:definition class="compliance" id="oval:ssg-grub2_password:def:1" version="1">
          <oval-def:metadata>
            <oval-def:title>Set Boot Loader Password in grub2</oval-def:title>
            <oval-def:affected family="unix">
              <oval-def:platform>Red Hat Enterprise Linux 8</oval-def:platform>
            </oval-def:affected>
            <oval-def:description>The grub2 boot loader should have password protection enabled.</oval-def:description>
            <oval-def:reference ref_id="CCE-80828-7" source="CCE"/>
            <oval-def:reference ref_id="grub2_password" source="ssg"/>
          </oval-def:metadata>
          <oval-def:criteria operator="OR">
            <oval-def:criterion comment="Pass if there are no files matching pattern '/boot/grub2/grub.cfg' exist in the system" test_ref="oval:ssg-test_grub2_password_file_boot_grub2_grub_cfg_absent:tst:1"/>
            <oval-def:criteria operator="AND">
              <oval-def:criteria comment="check both files to account for procedure change in documenation" operator="OR">
                <oval-def:criterion comment="make sure a password is defined in /boot/grub2/user.cfg" test_ref="oval:ssg-test_grub2_password_usercfg:tst:1"/>
                <oval-def:criterion comment="make sure a password is defined in /boot/grub2/grub.cfg" test_ref="oval:ssg-test_grub2_password_grubcfg:tst:1"/>
              </oval-def:criteria>
              <oval-def:criterion comment="make sure a superuser is defined in /boot/grub2/grub.cfg" test_ref="oval:ssg-test_bootloader_superuser:tst:1"/>
            </oval-def:criteria>
          </oval-def:criteria>
        </oval-def:definition>
```
Title and description aren't part of criteria and this displaying can be confusing, but I think they are useful information. If the second node (top criteria node) has any comment it doesn't display description but comment. 
I hope I am answered about the problem.

ggbecker:
Okay, if that's intended I'm ok with that. I haven't really spent too much time looking into this. I hope the fix proposed here is enough then. Thanks for the thorough analysis.

PR #164, opened 2020-12-22
Fix imports

Honny1:
This PR fixes import sorting and removes unused imports.


matejak:
Nice housekeeping, I like the new style. Merging.

PR #171, opened 2021-02-22
Test script for testing command entry points

Honny1:
This PR creates a tests suite for testing entry points installed by `rpm` or `pip`.


test_commands.sh
```diff
@@ -0,0 +1,212 @@
...
+-h    | --help                   Brings up this menu
+EOF
+}
+
+clean=false
+verbose=false
+silent=false
+oval_graph_install=false
+
+for arg in "$@"; do
```
matejak, line 28: I see that you parse command-line arguments manually. In your case, you could consider using a code generator to generate the help message and argument parsing code for you (e.g. [Argbash](https://argbash.readthedocs.org)), or you could consider using `getopts` to avoid the for loop.

test_commands.sh
```diff
@@ -0,0 +1,212 @@
...
+        exit 1
+    fi
+}
+
+install_package_from_source() {
+    install_pip="pip install --upgrade pip"
+    install_oval_graph="pip install -e ."
+
+    if [ $silent = false ] && [ $verbose = false ]; then
+        echo "Start install"
```
matejak, line 79: If verbose is false, then I think that these echos shouldn't be here.

test_commands.sh
```diff
@@ -0,0 +1,212 @@
...
+        $install_pip >/dev/null
+        $install_oval_graph >/dev/null
+        echo "End install"
+    elif [ $verbose = true ] && [ $silent = false ]; then
+        echo "Start install"
+        install_pip
+        install_oval_graph
+        echo "End install"
+    else
+        install_pip >/dev/null 2>&1
```
matejak, line 89: Earlier, you use the `$install_pip >/dev/null` redirection, and here you use `install_pip >/dev/null 2>&1`, which is inconsistent. The `&>` redirection (that works only in Bash) redirects both stdout and stderr.

test_commands.sh
```diff
@@ -0,0 +1,212 @@
...
+        install_oval_graph >/dev/null 2>&1
+    fi
+}
+
+report() {
+    if [ $test_result -eq 0 ]; then
+        printf "Result: %-70s \x1b[32mpassed\x1b[0m\n" "$*"
```
matejak, line 96: By convention, if everything works as expected, you don't report that, unless verbose output is requested.

matejak (changes requested):
There are some inconsistencies in the script that are mainly caused by necessity to handle combination of verbose on/off and silent on/off. I would suggest to get rid of any option that is not needed, as shells don't offer any reasonable possibility to resolve this without code duplication and copy-pasting.

matejak:
LGTM, failed packit jobs are unrelated to this PR.

PR #173, opened 2021-03-03
Rework oval_node

Honny1:
This PR contains:

- create docstrings 
- rework `OvalNode` class
- rework OVAL evaluation of operators
- rework `Converter` class
- rename and rework `_builder_oval_tree` class 
- Creates sup package `oval_tree` which contains classes that represent an OVAL tree and work with OVAL tree.


TomasTomecek:
@Honny1 hey, we're aware of the build failures and are tracking the problem here: https://github.com/packit/packit-service/issues/1016

Sorry about this, I hope we'd able to track the problem down by the end of the week.

Honny1:
@TomasTomecek Thanks for the info.

TomasTomecek:
I'm really puzzled why the build here fails the way it fails. I'll try to take a look into it.

Honny1:
@TomasTomecek Thanks!  
I figured out how to avoid this problem. Removing ``current_version_command: ["python3", "setup.py", "--version"]``  from the packit config, but this will not solve the problem of getting the version using the ``python3 setup.py --version`` command.

oval_graph/oval_node.py
```diff
@@ -19,9 +19,7 @@ class OvalNode():
     """
 
     def __init__(self, **kwargs):
```
matejak, line 22: Just curious - why the function accepts only kwargs, wile the comment suggests that there are required arguments?

oval_graph/oval_node.py
```diff
@@ -56,20 +57,23 @@ def __init__(self, **kwargs):
...
         if node_type not in ("value", "operator"):
             raise TypeError("Wrong value of node_type argument!")
         return node_type
 
-    def _validate_value(self, input_value):
+    @staticmethod
+    def _validate_value(input_node_type, input_value):
         value = input_value.lower()
 
         allowed_values = [
```
matejak, line 80: Just a note - those two data structures could be global sets (instead of local lists). The "in" operator works faster with sets.

oval_graph/oval_tree/oval_node.py
```diff
@@ -1,4 +1,5 @@
-import oval_graph.evaluate
+from .evaluate import (is_notapp_result, oval_operator_and, oval_operator_one,
```
matejak, line 1: I would use `from . import evaluate` - it's quite easy to write, and it also puts only one symbol to the global namespace.

oval_graph/oval_tree/evaluate.py
```diff
@@ -44,10 +68,21 @@ def oval_operator_one(result):
...
-    """
-        The OR operator produces a true result if one or more arguments is true. If every argument
-        is false, the result of the OR is false. If one or more of the arguments are unknown and
-        if none of arguments are true, then the OR operator produces a result of unknown.
+    """The OR operator produces a true result if one or more arguments is true. If every argument
+       is false, the result of the OR is false. If one or more of the arguments are unknown and
+       if none of arguments are true, then the OR operator produces a result of unknown.
+
+    Args:
+            result (dict): Dictionary representing frequencies of values
```
matejak, line 74: You probably mean "Dictionary storing occurrence of given values". Frequency has a slightly different meaning. This applies to other "frequencies" docstrings as well.

oval_graph/oval_tree/evaluate.py
```diff
@@ -144,26 +219,58 @@ def one_is_unknown(result):
...
+    Returns:
+            bool.
+    """
     return (eq_zero(result, 'true_cnt')
             and eq_or_greater_zero(result, 'false_cnt')
             and error_unknown_noteval_eq_zero(result)
             and result['notappl_cnt'] >= 0)
 
 
 def one_is_false1(result):
```
matejak, line 243: What's the difference between this and the `one_is_false` one?

oval_graph/oval_tree/evaluate.py
```diff
@@ -44,10 +68,21 @@ def oval_operator_one(result):
...
+            str. return values::
+
+                true
+                false
+                error
+                unknown
+                noteval
     """
     out_result = None
     if greater_zero(result, 'true_cnt'):
```
matejak, line 88: `true_cnt` etc. are not good names, it's better to use at least `true_count` or `num_of_true` (num is an established abbreviation, cnt is not).

oval_graph/oval_tree/evaluate.py
```diff
@@ -113,23 +111,18 @@ def oval_operator_xor(result):
                 noteval
     """
     out_result = None
-    if (result['true_cnt'] % 2) == 1\
-            and eq_zero_unknown_noteval_notappl(result):
+    eval_error_cnt = eq_zero(result, 'error_cnt')
```
matejak, line 114: It's not easy to see how `eq_zero(result, 'error_cnt')` is better than `result.num_errors == 0` (that would work if result is a named tuple) or `result.no_errors()` (that one would require result to be a class)

matejak (changes requested):
I have identified aspects of this PR that can be improved upon. At the same time, other things that I have noticed could be addressed in another PR.

oval_graph/oval_tree/evaluate.py
Honny1: Operator ONE has two cases of false state. I merge it into one function.

oval_graph/oval_tree/evaluate.py
Honny1: I create a results class and change bad naming in another PR. I create an Issue about it.  #192

matejak:
LGTM, thanks!

matejak (approved):

PR #177, opened 2021-03-16
Rework of HTML creation

Honny1:
This PR contains:
* Moves class for generating HTML to separate directory and renames it
* Rework generating HTML
* Changes how are opening reports with a web browser.
* Changes verbose output for generating files.


oval_graph/html_builder/graph.py
```diff
@@ -19,13 +19,10 @@ def __init__(self, parts, verbose, all_in_one):
...
-    def save_html(self, dict_oval_trees, src, rules):
-        self.save_html_report(dict_oval_trees, src)
-        self.print_output_message(src, self._format_rules_output(rules))
-
-    def save_html_report(self, dict_of_rules, src):
+    def save_html(self, dict_oval_trees, src):
         with open(src, "wb+") as data_file:
-            data_file.writelines(self._get_html(dict_of_rules))
+            data_file.writelines(self._get_html(dict_oval_trees))
+        self.print_output_message(src, list(dict_oval_trees.keys()))
```
matejak, line 24: I think that it is better to have the "if verbose" condition here than inside `print_output_message`. The disadvantage of this is that although you call `print_output_message(...)`, nothing is printed if the verbose mode is off, and it is not obvious why.

oval_graph/command_line_client/client_html_output.py
```diff
@@ -27,9 +27,8 @@ def get_src(src):
         return str(os.path.join(_dir, src))
 
     def prepare_data(self, rules):
-        out_src = []
         oval_tree_dict = dict()
-        self._prepare_data(rules, oval_tree_dict, out_src)
+        out_src = self._prepare_data(rules, oval_tree_dict)
```
matejak: `out_src` is a bad name, one can't guess what it represents.

oval_graph/html_builder/graph.py
```diff
@@ -81,27 +79,24 @@ def _get_html_body(self, dict_of_rules):
...
-                r'[\_\-\.]',
-                '',
-                k): v for k,
-            v in dict_of_rules.items()}
-        return ("var data_of_tree = " +
-                str(json.dumps(json_of_graphs, sort_keys=False, indent=4)) +
-                ";")
+    @staticmethod
+    def _remove_unfit_chars(string):
+        return re.sub(r'[\_\-\.]', '', string)
```
matejak, line 84: You don't have to escape `_` or `-`.

matejak (approved):

matejak:
LGTM, thanks for those fixes!

PR #181, opened 2021-03-17
Fix CWE-628

Honny1:
This PR fixes CWE-628.


matejak:
The CWE is quite vague, so let's hope that it got fixed. I don't observe any changes in behavior, so I merge it. Good job investigating this!

PR #185, opened 2021-03-31
Fix CVE-2021-28957

Honny1:
This PR fixes CVE-2021-28957.


Honny1:
Was resolved in #186.

PR #187, opened 2021-04-06
Rework of ARF XML parser

Honny1:
This PR contains:

* Creates sup package ```arf_xml_parser``` which contains classes parsing ARF XML.
* Reworks classes for parsing ARF XML.
  * Renames classes for parsing ARF XML.
  * Adds to the XML Parser the ability to load rules whose results are `noteval` and `notappl`. However, these rules cannot be visualized.
  * Changes methods to static methods.
  * Creates global namespaces for parsing XML.
* Changes how are loaded path


matejak:
Thank you for the PR! First thing - PR description: Please try to be more specific than "rework", because at least 80% of any software engineering falls under "rework".

I like how you extracted functionality into separate methods and made some methods static.
On the other hand, there is room for improvement in naming, and I think that parts of the code aren't unit-tested, which would be a pity.

Finally, don't squash commits right away, the review is easier when I can see what changed, and you can then rebase and create a nice history after everything is approved.

oval_graph/arf_xml_parser/arf_xml_parser.py
```diff
@@ -0,0 +1,117 @@
...
+            self.oval_definitions = self._get_oval_definitions()
+            self.scan_definitions = _OVALScanDefinitions(
+                self.definitions, self.oval_definitions, self.report_data).get_scan()
+        except BaseException as error:
+            raise ValueError(
+                'This file "{}" is not arf report file or there are no results'.format(
+                    self.src)) from error
+
+    @staticmethod
+    def get_src(src):
```
matejak, line 42: You could use [pathlib](https://docs.python.org/3/library/pathlib.html) for dealing with paths, that way you can just declare the local data directory as a `pathlib.Path`, and then do things like `xsd_path = LOCAL_DATADIR / xsd_basename` etc.
`get_src` is not a very good name, and pathlib does what is being done there.

matejak (changes requested):
Majority of suggestions are related to naming - the PR itself leads to the improved structure of the code.

oval_graph/arf_xml_parser/arf_xml_parser.py
```diff
@@ -12,14 +12,16 @@
 from ._oval_scan_definitions import _OVALScanDefinitions
 from .global_namespaces import namespaces
 
+LOCAL_DATA_DIR = Path(__file__).parent.joinpath('../schemas/')
```
matejak, line 15: Don't use `joinpath` - the `/` operator does the same and is more elegant, i.e. `Path(__file__).parent.parent / "schemas"`

matejak (approved):

matejak:
LGTM, thanks!

PR #188, opened 2021-04-26
test_commands.sh: always show test and its result

matusmarhefka:
This will show test name and its result all the time not only when using verbose mode (which shows also command outputs).

### DIFF
--- a/test_commands.sh
+++ b/test_commands.sh
@@ -135,12 +135,10 @@ install_package_from_source() {
 }
 
 report() {
-    if [ "$_arg_verbose" = "on" ]; then
-        if [ $test_result -eq 0 ]; then
-            printf "Result: %-70s \x1b[32mpassed\x1b[0m\n" "$*"
-        else
-            printf "Result: %-70s \x1b[31mfailed\x1b[0m\n\n" "$*"
-        fi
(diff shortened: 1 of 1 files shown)


matejak:
Is the aim to show results of all tests, or just of failed ones?
The Unix philosophy advocates that programs should not report by default when subtasks go well according to expectations. In this context, I see that the failure should be reported always, but do we need the passing test report outside of the verbose mode?

matusmarhefka:
> Is the aim to show results of all tests, or just of failed ones?
> The Unix philosophy advocates that programs should not report by default when subtasks go well according to expectations. In this context, I see that the failure should be reported always, but do we need the passing test report outside of the verbose mode?

Right now the test script prints only a version of `oval-graph` in non-verbose mode so it is not really clear what is being tested. The idea was to see what tests are being executed and their results (even passing ones) without additional verbosity produced by the verbose mode. But I don't have a strong opinion about it.

Honny1:
Hello @matusmarhefka and @matejak, 
I suggest closing this PR and creating an issue to change the output of ```test_commands.sh```. I think it will be good to show the results only when something is wrong and to hide the ```oval-graph``` version. Then the ```test_commands.sh``` script would create a log file with the complete result.

PR #189, opened 2021-04-27
Make sure that users can run tests from the sdist.

matejak:
Test code, support scripts and data was not packaged along, so people who downloaded the pypi source package (that is also used as a Fedora source) can run the whole test suite without having to check out those files from Git.


Honny1:
I like it, merging!

PR #191, opened 2021-05-21
The output of test_commands.sh

### LINKED ISSUE #190: Change output of the test_commands.sh
Modify `test_command.sh` to show results only when something is wrong and remove showing of the `oval-graph` version. Then the `test_commands.sh` script would create a log file with the complete result.

Honny1:
This PR changes the output of `test_commands.sh` according to #190.

Fixes #190

### DIFF
--- a/test_commands.sh
+++ b/test_commands.sh
@@ -142,7 +142,7 @@ report() {
     fi
 }
 
-test() {
+test_command() {
     test_name="$1"
     command="$2"
     msg=""
@@ -161,7 +161,7 @@ test() {
     report "${msg}"
 }
 
-test_rise_error() {
+test_command_rise_error() {
     test_name="$1"
     command="$2"
     msg=""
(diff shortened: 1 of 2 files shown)


test_commands.sh
```diff
@@ -195,39 +189,46 @@ clean() {
...
 hide_all_passing_tests_test() {
-    test run-arf-to-graph "arf-to-graph -o ${tmp_dir_src} ${test_file_src} --hide-passing-tests fips"
-    test run-json-to-graph "json-to-graph -o ${tmp_dir_src} ${tmp_json_file_src} --hide-passing-tests -i fips"
+    test_command run-arf-to-graph "arf-to-graph -o ${tmp_dir_src} ${test_file_src} --hide-passing-tests fips"
+    test_command run-json-to-graph "json-to-graph -o ${tmp_dir_src} ${tmp_json_file_src} --hide-passing-tests -i fips"
 }
 
+
+# Backup descriptors
+exec 3>&1 4>&2
```
matejak, line 227: You have used an advanced Bash feature: https://copyconstruct.medium.com/bash-redirection-fun-with-descriptors-e799ec5a3c16

test_commands.sh
```diff
@@ -135,50 +135,44 @@ install_package_from_source() {
...
 
 report() {
-    if [ "$_arg_verbose" = "on" ]; then
-        if [ $test_result -eq 0 ]; then
-            printf "Result: %-70s \x1b[32mpassed\x1b[0m\n" "$*"
-        else
-            printf "Result: %-70s \x1b[31mfailed\x1b[0m\n\n" "$*"
-        fi
+    if [ $test_result -eq 0 ]; then
+        printf "Result: %-70s \x1b[32mpassed\x1b[0m\n" "$*"
```
matejak, line 139: Argbash will complain about unbalanced square brackets, which can be fixed by a comment:

```suggestion
        printf "Result: %-70s \x1b[32mpassed\x1b[0m\n" "$*"  # ]]
```

matejak:
I would probably choose a different approach using e.g. `log_verbose` and `log_always` functions, when there would be little of output if everything works, but errors would appear every time, although maybe not is full detail.
However, that is just a matter of taste. I don't have any problem with this PR, but I would ask @matusmarhefka for an opinion, and if he is not against it, for taking over it.

matusmarhefka (approved):
LGTM, I actually like this approach.

PR #198, opened 2021-06-02
Rework OVAL gate evaluation

### LINKED ISSUE #192: Improve storing the result of the OVAL tree
It will be a good idea to store the tree result in a different structure than the dictionary. Rename the items in the new structure to make them easier to understand. For example, `cnt_true` to `number_of_true`.

Honny1:
This PR reworks OVAL gate evaluation.

Fixes #192


matejak (approved):
The transformation of the result format is well done, and I believe that it will be easier to work with.

PR #202, opened 2021-06-24
Rework tests for command line 

Honny1:
This PR is the next part of the rework of the test suite. This rework is focused on tests for commands of the oval-graph package.


tests_oval_graph/test_tools.py
```diff
@@ -0,0 +1,95 @@
...
+                matched = True
+                data_in_html = row
+        assert matched
+
+        tmp_json_str = data_in_html.replace(prefix_start, '').replace(prefix_end, '')
+        tmp_json = json.loads(tmp_json_str)
+        data_in_html = prefix_start + json.dumps(tmp_json, indent=4, sort_keys=False) + prefix_end
+
+        count_row = 0
+        rule_name = 'xccdforgssgprojectcontentrulepackageabrtremoved'
```
jan-cerny, line 54: What is this?

tests_oval_graph/test_tools.py
```diff
@@ -0,0 +1,95 @@
...
+        reference_pattern = TestTools.get_text_file(
+            'test_commands/test_data/referenc_pattern_html_report.txt')
+        prefix_start = '<script>var data_of_tree = '
+        prefix_end = ';</script><div>\n'
+        data_in_html = ""
+        matched = False
+        for row in result_:
+            if prefix_start in row and prefix_end in row:
+                matched = True
+                data_in_html = row
```
jan-cerny, line 46: I I understand it well you can have a `break` statement here.

tests_oval_graph/test_commands/test_command_bad_arf.py
```diff
@@ -0,0 +1,59 @@
...
+import pytest
+
+COMMAND_START = ['python3',
+                 '-m',
+                 'oval_graph.command_line'
+                 ]
+
+JSON_TO_GRAPH_COMMAND = [*COMMAND_START,
+                         'json-to-graph',
+                         'tests_oval_graph/global_test_data/ssg-fedora-ds-arf.xml',
```
jan-cerny, line 12: I would define `tests_oval_graph/global_test_data/` as a constant

jan-cerny (commented):
It looks good to me, I have only minor comments.

tests_oval_graph/test_tools.py
Honny1: It is the ID of the rule in the HTML report. Because jQuery has a problem with non-alphabet characters.

PR #203, opened 2021-06-24
Rework tests for command line client

Honny1:
This PR is the next part of the rework of the test suite. This rework is focused on tests for the command line client.


tests_oval_graph/test_commands/test_command_line_client/test_client_arf_to_html.py
```diff
@@ -0,0 +1,76 @@
...
+    rules = {'rules': [rule]}
+    with pytest.raises(Exception, match=error_pattern):
+        assert client.prepare_data(rules)
+
+
+@pytest.mark.parametrize("rule, error_pattern", SEARCH_BAD_RULES)
+def test_expection_search_rules_id(rule, error_pattern):
+    client = get_client_arf_to_html(rule)
+    with pytest.raises(Exception, match=error_pattern):
+        assert client.search_rules_id()
```
jan-cerny, line 39: Could we have a test that tests the `ValueError` and `NotTestedRule` exceptions raised at `oval_graph/command_line_client/client_arf_input.py` lines 56 or 58 ?

tests_oval_graph/test_commands/test_command_line_client/test_client_json_to_html.py
```diff
@@ -0,0 +1,75 @@
...
+        args.extend(optional_args)
+    client = JsonToHtml(args)
+    client.load_file()
+    return client
+
+
+@pytest.mark.parametrize("src, error_pattern", [
+    ('../../global_test_data/ssg-fedora-ds-arf.xml', 'is not valid json'),
+    ('../test_data/referenc_html_report.html', 'No such file or directory:'),
+    ('../test_data/bad_result_data_json.json', 'valid for OVAL tree')
```
jan-cerny, line 29: The paths to these files repeat multiple times and in multiple files across this PR. Could we define them on a single place? Would there be a way to avoid the relative paths (the two dots `..`)?

PR #210, opened 2021-08-11
Fix problem with saving the report to dir or to file when is used `--output`

Honny1:
This PR fixes problem with saving the report to dir or to file when is used `--output`. When the user wants to generate more reports it will create OUTPUT as a directory.

### DIFF
--- a/oval_graph/command_line_client/client.py
+++ b/oval_graph/command_line_client/client.py
@@ -204,7 +204,8 @@ class Client():
             '--output',
             action="store",
             default=None,
-            help='The file where to save output.')
+            help=("The path where to save the output. If there are more report "
+                  "files to generate it will create an OUTPUT path as a directory."))
         parser.add_argument(
             "source_filename",
             help=self._get_message().get('source_filename'))
--- a/oval_graph/command_line_client/client_html_output.py
+++ b/oval_graph/command_line_client/client_html_output.py
@@ -1,7 +1,7 @@
-import os
 import tempfile
 import time
 import webbrowser
+from pathlib import Path
 from subprocess import PIPE, Popen, check_call
 
 from ..exceptions import NotTestedRule
@@ -20,6 +20,7 @@ class ClientHtmlOutput(Client):
         self.display_html = True if self.out is None else self.arg.display
         self.html_builder = Graph(self.arg.verbose, self.all_in_one)
         self.web_browsers = []
+        self.selected_only_one_rule = False
 
     def prepare_data(self, rules):
         paths_to_generated_rules = self._prepare_data(rules)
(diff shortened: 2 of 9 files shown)


oval_graph/command_line_client/client.py
```diff
@@ -204,7 +204,8 @@ def prepare_parser(self, parser):
             '--output',
             action="store",
             default=None,
-            help='The file where to save output.')
+            help=("The path where to save the output. If there is more report"
+                  " to generate it will create an OUTPUT path as a directory."))
```
matejak, line 208:
```suggestion
            help=("The path where to save the output. If there are more report "
                  "files to generate it will create an OUTPUT path as a directory."))
```

oval_graph/command_line_client/client_html_output.py
```diff
@@ -20,6 +20,7 @@ def __init__(self, args):
         self.display_html = True if self.out is None else self.arg.display
         self.html_builder = Graph(self.arg.verbose, self.all_in_one)
         self.web_browsers = []
+        self.is_chosen_more_rules = False
```
matejak, line 23: probably `selected_multiple_rules` would be a better name, but because you use it in a negation, why not to have `selected_only_one_rule` instead?

oval_graph/command_line_client/client_html_output.py
```diff
@@ -64,6 +67,9 @@ def get_file_name(rule):
 
     def get_save_src(self, rule):
```
matejak, line 68: What could be "save_src"? It looks like that the method returns the filename where to save the output data. Apart from that, the method can create directories, so it's not only getting filenames, but it prepares their directories as well.

The fact that the method modifies the system also affects your ability to test it, I recommend to just output the path, and you can use `pathlib`, e.g. `p.parent.mkdir(parents=True, exist_ok=True)`

tests_oval_graph/test_commands/test_command_arf_to_graph.py
```diff
@@ -94,4 +91,4 @@ def test_command_parameter_all_and_show_failed_rules():
                r'_package_\w+_removed'
                ]
     subprocess.check_call(command)
-    assert len(os.listdir(src)) == 1
+    assert os.path.isfile(src)
```
matejak, line 94: How come that the output of `get_random_dir...` can be a file and not a directory? Perhaps that function should be renamed to refer to `path` instead of `dir`

Next, I would recommend to use `pathlib` everywhere, and then you can write just `src.is_file()` - much shorter!

tests_oval_graph/test_commands/test_command_json_to_graph.py
```diff
@@ -111,9 +110,7 @@ def test_inquirer_choice_rule():
     for key_ in keys:
         sut.send(key_)
     sut.wait()
-    assert len(os.listdir(out_dir)) == 1
-    assert ("xccdf_org.ssgproject.content_rule_package_abrt_removed"
-            in os.listdir(out_dir)[0])
+    assert os.path.isfile(out_dir)
```
matejak, line 113: `out_dir` is file? Then we have another naming issue :-)

matejak (approved):
The PR works great, but attention to detail is important for long-term well-being of the project and its contributors. Numerous naming issues have surfaced that I recommend to take action upon.

matejak:
LGTM, the test failure is purely infrastructure-related. Good job!

